Issue #12, opened 2022-02-21
Add flag in RegisteredZone for `SupportsMultiSend`


joe-bowman:
When we submit a MsgRegisterZone tx, we need to specify whether the zone we are connecting to permits the sending of MsgMultiSend txs - if not, we can fall back to using MsgSend, but have to send many of these and it is less efficient.

RegisterdZone already implements `func SupportsMultiSend() bool` but this is currently hardcoded to `false`.

---
Acceptance Criteria:
- [x] MsgRegisterZone tx includes boolean `support_multi_send` field
- [x] `quicksilverd tx interchainstaking register-zone` CLI has a flag for MultiSend support from CLI

joe-bowman:
This change requires:
 - a new boolean field added to the RegisteredZone protobuf message in `interchainstaking/v1/genesis.proto`
 - regenerate proto files (`make gen-proto`)
 - update `NewMsgRegisterZone` in `x/interchainstaking/types/msgs.go` to set the new flag
 - add a new optional argument to `GetRegisterZoneTxCmd` in `x/interchainstaking/client/cli/tx.go` (default: false) and have it passed into the tx
 - update the `scripts/simple-test.sh` call to `quicksilverd tx interchainstaking register-zone` to pass the new flag (with value = true; as we explicitly enable `MsgMultiSend` in the genesis of testzone.

joe-bowman:
There is also a method in `x/interchainstaking/types/zones.go` called `SupportMultiSend()` which currently is hard coded to return `false` but need to be updated to return the value of the new field.

PR #35, opened 2022-03-07
issue #12: multi send support

zanicar:
- added multi_send param to RegisteredZone (protobuf)
- added multi_send param to MsgRegisteredZone (protobuf)
- updated function signature NewMsgRegisteredZone with multi_send param
- updated SupportMultiSend method to return MultiSend property
- added min selector on delegationAccounts
- updated script to use multi-send param on interchainstaking


x/interchainstaking/types/zones.go
```diff
@@ -14,12 +15,12 @@ func (z RegisteredZone) GetDelegationAccountsByLowestBalance(qty int64) []*ICAAc
 		return delegationAccounts[i].DelegatedBalance.Amount.GT(delegationAccounts[j].DelegatedBalance.Amount)
 	})
 	if qty > 0 {
-		return delegationAccounts[:qty]
+		return delegationAccounts[:int(math.Min(float64(len(delegationAccounts)-1), float64(qty)))]
```
joe-bowman, line 18: 👍🏻

joe-bowman (changes requested):
need rebasing on the latest push to reward-withdrawals but otherwise GTG

PR #43, opened 2022-03-14
Ibc v3.0.0 rc2

### LINKED ISSUE #42: Handle ICA Acknowledgements
We should be using IBC Acknowledgement packets (and similarly timeout packets) to prove X has occurred, rather than relying on changing balances. This wasn't necessary for Delegations but is required to ensure we can correctly match withdrawal amounts.

joe-bowman:
- bump ibc-go to v3.0.0-rc2
- add stub handlers for dealing with acknowledgements (closes #42)


zanicar (approved):
LGTM

PR #45, opened 2022-03-14
Fix intent bug

joe-bowman:
Fix bug in intent logic where we don't escape enough loops after updating an existing record, so we end up with duplicate entries. 

Also add tests for the error case.


zanicar (approved):
LGTM

Issue #38, opened 2022-03-07
Fix output of `make lint`


joe-bowman:
Running `make lint` returns the following errors:
```
x/interchainquery/keeper/msg_server.go:30:22    errcheck  Error return value of `k.SetDatapointForId` is not checked
x/interchainquery/keeper/msg_server.go:36:23    errcheck  Error return value of `k.SetDatapointForId` is not checked
x/interchainstaking/keeper/abci.go:134:16       errcheck  Error return value of `k.Delegate` is not checked
x/interchainstaking/keeper/hooks.go:19:32       errcheck  Error return value of `k.WithdrawDelegationRewards` is not checked
x/interchainstaking/keeper/msg_server.go:38:24  errcheck  Error return value of `k.SetConnectionForPort` is not checked
x/interchainstaking/keeper/msg_server.go:47:25  errcheck  Error return value of `k.SetConnectionForPort` is not checked
x/interchainstaking/keeper/receipt.go:98:14     errcheck  Error return value of `k.MintQAsset` is not checked
x/interchainstaking/keeper/receipt.go:99:22     errcheck  Error return value of `k.TransferToDelegate` is not checked
app/tps_counter_test.go:26:14                   errcheck  Error return value of `tpc.start` is not checked
```

levicook:
my draft pull request cleans up all `errcheck` complaints -- returns the error where it's compatible to; logs it when it can't.

levicook:
There's one error left on the pull request that appears to run a bit deeper; looks like fixing this requires a package upgrade that isn't fully compatible. Not sure on the ripple here & thinking it probably belongs in a different pull request, so just going to submit the PR with this still failing:

```
make lint
golangci-lint run --out-format=tab
x/interchainstaking/keeper/ibc_handlers.go:6:2  staticcheck  SA1019: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead. 
make: *** [lint] Error 1
```

levicook:
Should be able to close this after that last PR is merged.

joe-bowman:
closed by #51 #53

PR #53, opened 2022-04-01
Fix output of `make lint`, take 2 of 2

levicook:
Last tiny fix for https://github.com/ingenuity-build/quicksilver/issues/38


joe-bowman:
In theory we should scope this to the error code: e.g. 

`//lint:ignore SA1004 ignore this!`

levicook:
Ah, that's even better. Had a hard time finding the _right_ way to convince golangci-lint to ignore this. Will update to match yours.

joe-bowman (approved):
lgtm

PR #56, opened 2022-04-06
Add redemption workflow

### LINKED ISSUE #14: Exit workflow
As a: user
I want: to submit a MsgRedeemDelegation transaction
So that: I can redeem my qAsset tokens for the originally staked collateral

### LINKED ISSUE #34: Cache aggregatedIntent in RegisteredZone
User intent is updated on a sporadic basis, based upon MsgSignalIntent and also for inbound delgations. 

Updating is relatively cheap because in each instance only a single users' delegation is fetched and stored.

Aggregating a zone's intent however, causing many users' Intents to be fetched and many account balances to be queried. The size of the dataset here is unbounded so we should aim to fetch / aggregate this as infrequently as possible.

There is also an potential 'attack' here whereby a user could change their Intent immediately before epoch in order to ensure they are signalled to the _best_ validators for that epoch.

As such we should conduct this aggregation epochly, immediately AFTER participation reward calculations, and immediately BEFORE before undertaking (re)delegations.

This value can be cached in the RegisteredZone object, as it will be used frequently enough, and is off small enough size to jusitfy it.

### LINKED ISSUE #50: Bug: epoch ordering
At the end of the epoch, we call `endEpoch(epochInfo)` _before_ we all AfterEndEpoch with the _current_ (thus, already incremented) epoch height.

We should move `endEpoch()` (which just increments the epochRecord) to be the last thing - after calling AfterEndEpoch (as unintuitive as this sounds).

joe-bowman:
- add redemption workflow (resolves #14)
- fix ordering in epoch end blocker (resolves #50)
- bump ibc-go to 3.0.0
- bump cosmos-sdk to v0.45.1-ls-2
- cache intents epochly (resolves #34)
- fetch outstanding reward data via ICQ
- handle acknowledgements

### DIFF
--- a/x/interchainstaking/keeper/keeper.go
+++ b/x/interchainstaking/keeper/keeper.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/store/prefix"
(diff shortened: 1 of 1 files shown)


zanicar (approved):
LGTM

Issue #59, opened 2022-04-06
Spec: FeeAccount


joe-bowman:
- Add a new *ICAAccount to RegisteredZone type to handle fee collection
- initialize on msg_server.go MsgRegisterZone handler
- handle OnChanOpenAck callback in ibc_module.go
- On distribution of rewards, `MsgSend` parmas.CommissionRate * balance -> FeeAccount.

Future:
- FeeAccount can send to DepositAddress for rebonding
- Deposits from FeeAccount are distributed to all QCK stakers (depends on Staking / Governance logic).

joe-bowman:
This was added by #65, but can now be removed in light of ibc tx directly from withdrawal.

Issue #27, opened 2022-03-02
Use uint64 instead of sdk.Int when dealing with blockheights in Interchain Queries


joe-bowman:
Currently Interchain queries uses sdk.Int throughout (as this matches how block height is stored within `ctx`), however, we spend most of our time marshalling between sdk.Int andunit64 anyway, so lets make our lives easier :)

PR #63, opened 2022-04-07
add params to interchain staking module

joe-bowman:
- add parameter for existing constant values
- add commission_rate param ready for #59


zanicar (approved):
LGTM

zanicar:
Looks like it also solves #27...
or only partially?

joe-bowman:
> Looks like it also solves #27...
> or only partially?

Ish. There is a whole bunch of other stuff in there that used sdk.Int sadly.

Issue #61, opened 2022-04-06
Onboarding qAssets should be calculated using `last_redemption_rate`


joe-bowman:
In order to circumvent potential attack described here: https://discord.com/channels/939972129699561472/939973146558218291/960210775182561320

We should mint new qAssets at the `min()` of the current `redemption_rate` and the `redemption_rate` of the previous epoch (held in `RegisteredZone.LastRedemptionRate` - see #60). For long term qAsset holders, this should be a negligible impact, but serves to penalises users who enter and then immediately exit the protocol.

zanicar:
Should it not mint qAssets at the current `redemption_rate` and burn the `min()` of the `redemption_rate` and the `last_redemption_rate`?

Meaning users will enter the protocol at the current rate and exit at the lesser of the current and previous rate... thereby preventing epoch boundary gaming, even in the event of slashing.

joe-bowman:
Yes, you are correct

joe-bowman:
partially closed by #64; blocked on WithdrawalAccount+FeeAccount piece.

joe-bowman:
closed by #70

Issue #60, opened 2022-04-06
Add `last_redemption_rate` to RegisteredZone


joe-bowman:
Add protobuf entries for `last_redemption_rate` of type `sdk.Dec`.

This will represent the redemption rate from the previous epoch. It should be set when the new redemption rate is calculated (it is set to the previous value).

joe-bowman:
closed by #64

PR #64, opened 2022-04-08
Add last_redemption_rate to RegisteredZone

zanicar:
Solves #60 and #61


joe-bowman (approved):
lgtm

joe-bowman (changes requested):
need to set the value of last_redemption_rate to 1.0 in msg_server (when zone is created)

zanicar:
Thanks for the catch, updated.
0.95 - don't think that is necessary, as epoch 0 will present a zero-sum game, so no incentive to attack.

joe-bowman:
> 0.95 - don't think that is necessary, as epoch 0 will present a zero-sum game, so no incentive to attack.

I think you are right. If anything the user loses out (zero rewards earned) and the pool benefits (earns rewards for that time period).

joe-bowman (approved):
lgtm

Issue #58, opened 2022-04-06
Spec: WithdrawalAccount


joe-bowman:
- Add a new *ICAAccount to `RegisteredZone` type to handle withdrawals
- initialize on msg_server.go MsgRegisterZone handler
- handle OnChanOpenAck callback in ibc_module.go
- on OnChanOpenAck for delegate accounts, send `MsgSetWithdrawalAddr` to set to WithdrawalAccount 

- Epochly, delegate accounts call MsgWithdrawDelegatorRewards (this already happens), and rewards are withdrawn to WithdrawalAccount
  - track 'outstanding' withdrawal txs (counter; inc when MsgWithdrawDelegatorRewards is sent; dec when Ack is received)
  - on MsgWithdrawDelegatorRewards acknowledgement, increment balance of withdrawal account
  - when tracking hits 0 (no outstanding withdrawals), sum balance, update redemption rate, and distribute to delegateAccounts.

- REMOVE PeriodicQueries for delegate account balance. Only MsgSend and MsgMultiSend (from a known account) should now be able to trigger a delegation (this can now come from Deposit OR Withdrawal)

zanicar:
Implemented proportional rewards distribution to delegation accounts;
Utilized option (b) fan out delegation pattern as it was already implemented in that way;

joe-bowman:
Closed by #65

PR #65, opened 2022-04-22
Zone accounts

zanicar:
Addresses #58, #59

TODO:
1. set query data point for withdrawal account balance;
2. update redemption rate - stub needs implementation;

Created stub for handleFee (see #59 Future)


joe-bowman (approved):
lgtm

x/interchainstaking/keeper/ibc_handlers.go
```diff
@@ -180,33 +181,84 @@ func (k *Keeper) HandleCompleteMultiSend(ctx sdk.Context, msg sdk.Msg) error {
...
 	if !ok {
-		k.Logger(ctx).Error("unable to cast source message to MsgSend")
-		return fmt.Errorf("unable to cast source message to MsgSend")
+		err := fmt.Errorf("unable to cast source message to MsgSend")
+		k.Logger(ctx).Error(err.Error())
+		return err
+	}
+
+	// get zone
+	zone, ok := k.GetRegisteredZoneInfo(ctx, ctx.ChainID())
```
joe-bowman, line 198: Just found this - posting here for posterity - `ctx.ChainID()` will always give the local chain ID. We need to do `k.GetZoneForDelegateAccount()` or similar here.

PR #66, opened 2022-04-23
Minor updates

joe-bowman:
Numerous minor update:
 - fix packetData unmarshalling bug (still need to fully understand _why_ the keeper's codec cannot unmarshal this correctly :/ )
 - fix swagger doc generation
 - update proto generation dependencies (v0.6)
 - add missing third_party protobuf defs (timestamp.proto and duration.proto)
 - lint and format proto messages
 - test var tweaks to ensure smooth running (override params to not generate 100 delegate accounts locally)
 - remove rocksdb from docker container


zanicar (approved):
LGTM

PR #67, opened 2022-04-27
Interchain Query Callbacks + Fee Collection

joe-bowman:
- merge ICQ Single Query and Periodic Query into single Query type (`period` of -1 == non-repeating)
- add ability to register callbacks against queries
- callbacks are executed on query return
- replace deposit accounts and validator status queries with callback functions
- add queries and callback for withdrawal account balance queries
- handle fee collection + ibc transfer of fees at epoch boundary


zanicar (approved):
LGTM

PR #68, opened 2022-04-28
IBC Context

zanicar:
Added some convenience methods that utilize context to store the current IBC connection id;

### DIFF
--- a/x/interchainstaking/ibc_module.go
+++ b/x/interchainstaking/ibc_module.go
@@ -1,6 +1,7 @@
 package interchainstaking
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


x/interchainstaking/ibc_module.go
```diff
@@ -217,6 +218,14 @@ func (im IBCModule) OnAcknowledgementPacket(
 	acknowledgement []byte,
 	relayer sdk.AccAddress,
 ) error {
+	connectionId, _, err := im.keeper.IBCKeeper.ChannelKeeper.GetChannelConnection(ctx, packet.SourcePort, packet.SourceChannel)
```
joe-bowman, line 221: is `packet` here the original packet being acknowledged (i.e. source is our local channel and port) - I think it is (in which case this is fine!).

x/interchainstaking/ibc_module.go
```diff
@@ -217,6 +218,14 @@ func (im IBCModule) OnAcknowledgementPacket(
...
+	if err != nil {
+		err = fmt.Errorf("packet connection not found: %w", err)
+		ctx.Logger().Error(err.Error())
+		return err
+	}
+	ctx = ctx.WithContext(context.WithValue(ctx.Context(), "connectionId", connectionId))
```
joe-bowman, line 227: Is it worth doing the same thing for `chain-id`?

joe-bowman:
This looks good but will need rebasing after #70

x/interchainstaking/ibc_module.go
zanicar: I was wondering the same thing...
but we are always able to get the chainId from the connectionId (convenience methods implemented in this PR),
and seeing as it is recommended to not add unnecessary values to context it seemed best to only add the conncetionId that would otherwise be unobtainable further down the execution stack;

x/interchainstaking/ibc_module.go
zanicar: I am a little confused here, but as you mention it seems to be the original packet, thus the source would refer to the local chain channel and port;

OnAcknowledgementPacket signature suggests packet is the original accompanied by acknowledgement []byte;

joe-bowman (approved):
lgtm

PR #70, opened 2022-04-29
Mint and redemption rate

joe-bowman:
- add forked mint module from osmosis
- update forked module to remove devRewards pool and add participationRewards pool
- change to epoch module and tests to accomodate mint module
- update delegations query to use callback
- update redemption rate
- add additional logging for callback failure


zanicar (approved):
LGTM

PR #74, opened 2022-05-01
use a default validator intent if the existing one is zero length

joe-bowman:
- add default validator intent (if zone aggregate intent is zero-length)
- equally across all validators with < 50% commission (this will need to be a param long term)
- add function to delegate any unexplained balance in the delegate account
- temporarily trigger the above fn from endblocker on DelegateInterval. Long term this will be triggered on delegate account creation.


zanicar (approved):
LGTM

PR #75, opened 2022-05-01
remove temporary delegate routine, and add as an icq callback

joe-bowman:
- remove temporary endblocker to settle outstanding delegations
- add interchainquery callback for settling outstanding delegations


zanicar (approved):
LGTM

PR #94, opened 2022-06-01
move delegation management into the KV store, and out of the zone object

joe-bowman:
- migrate delegations from under zone/validators to KVStore (memory + performance)
- fix deposit_address and delegator_intent queries
- add delegation queries (by zone, by delegator, by validator)
- use interchain queries kvstore lookup to validate delegation queries
- trigger valset lookup when ibc valset hash changes


zanicar (approved):
lgtm

PR #95, opened 2022-06-01
make memo field intent signalling use base64

joe-bowman:
- fix memo intent signalling to use base64
- add tests for `ConvertMemoToOrdinalIntents` and `ConvertCoinsToOrdinalIntents`


joe-bowman:
stacked diff - based on #94

zanicar (approved):
lgtm

PR #96, opened 2022-06-03
Participation rewards module

joe-bowman:


x/participationrewards/client/cli/query.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/ingenuity-build/quicksilver/x/mint/types"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+)
+
+// GetQueryCmd returns the cli query commands for the minting module.
+func GetQueryCmd() *cobra.Command {
+	mintingQueryCmd := &cobra.Command{
```
joe-bowman, line 16: copypasta from mint module :D ?

x/participationrewards/keeper/distribution.go
```diff
@@ -0,0 +1,74 @@
...
+	ValidatorSelection sdk.Coins
+	Holdings           sdk.Coins
+	Lockup             sdk.Coins
+}
+
+func (k Keeper) getRewardsAllocations(ctx sdk.Context) rewardsAllocation {
+	var allocation rewardsAllocation
+
+	moduleAddress := k.accountKeeper.GetModuleAddress(types.ModuleName)
+	moduleBalances := k.bankKeeper.GetAllBalances(ctx, moduleAddress)
```
joe-bowman, line 18: Given this is a known entity (will always be QCK), can we query just for the QCK balance of this account - saves the nasty `moduleBalances[0]` references below...

x/interchainstaking/ibc_module.go
```diff
@@ -209,6 +221,60 @@ func (im IBCModule) OnChanOpenAck(
...
+	}
+
+	var cb keeper.Callback = func(k keeper.Keeper, ctx sdk.Context, response []byte, query icqtypes.Query) error {
+		zone, found := k.GetRegisteredZoneInfo(ctx, query.GetChainId())
+		if !found {
+			return fmt.Errorf("no registered zone for chain id: %s", query.GetChainId())
+		}
+
+		// initialize performance delegations
+		if err := k.InitPerformanceDelegations(ctx, zone, response); err != nil {
```
joe-bowman, line 247: This is being called every keyDelegations blocks - is this correct?

zanicar: No, it should only run until successful... (initial performance delegations)

1. Query balance
2. If sufficient, send performance delegations...
3. Then, stop.

From here the zone callbacks dealing with QueryDelegationTotalRewards call WithdrawDelegatorReward for the performance account of the zone ("same as if the delegator simply started a new delegation of the same value"); This action resets the performance scoring for the next epoch...

We have to deal with changes to the validator set, such as validators dropping off (or joining) the active set, validators being jailed, etc; I think it would be best to hook performance delegation management mechanisms into that logic for better overall clarity.

x/participationrewards/keeper/distribution.go
zanicar: This allows for localized ibctesting, where DefaultTestingAppInit sets the base denom to ustake;
I can add another var `baseBalance := moduleBalances[0]` for code clarity if desired...

PR #97, opened 2022-06-04
Update delegation strategy

joe-bowman:
- add support for base64 memo for intent signalling.
- add better delegation planning
  - favour grouping delegations to the same validator together
  - persist delegation decision in receipt


zanicar (approved):
lgtm

PR #99, opened 2022-06-07
Add flag to indicate support for LSM module

### LINKED ISSUE #92: Add LSM flag to registered zones to conditionally allow deposit/withdrawals of delegated positions
We need to provide a flag on the RegisteredZone struct to indicate whether or not the zone has the LSM enabled, so we can conditionally enable/disable deposits of LSM tokens and withdrawals.

joe-bowman:
- Add flag to protobug definitions (struct and msg)
- Add flag setting to create registered zone 
- Add flag to cli tx
- Check flag on redemption

closes #92


zanicar (approved):
lgtm

PR #107, opened 2022-06-15
Fix participation panics

joe-bowman:
- revert ICQ height check introduced in #106 
- make memo -> intent errors clearer
- remove debug logging
- fix panics when nil allocations
- fix panics when nil rewards


zanicar (approved):
I made some changes to participation rewards and will PR for it soon... overriding some of these changes, but other than that LGTM :)

PR #111, opened 2022-06-16
participation rewards update

zanicar:
- added zone allocation state
- zero allocations on callback
- merged zone proportions and allocations (less looping required)
- devtest debug output removed
- removed unnecessary params from functions


x/participationrewards/keeper/callbacks.go
```diff
@@ -89,5 +85,9 @@ func RewardsCallback(k Keeper, ctx sdk.Context, response []byte, query icqtypes.
 		k.icsKeeper.SetIntent(ctx, zone, di, true)
 	}
 
+	// set zone ValidatorSelectionAllocation to zero
+	zone.ValidatorSelectionAllocation = sdk.NewCoins(sdk.NewCoin("uqck", sdk.ZeroInt()))
```
joe-bowman, line 89: whilst this always will be `uqck`, it's probably correct to fetch `k.BondDenom(ctx)` from the staking keeper.

x/participationrewards/keeper/rewards_validatorSelection.go
```diff
@@ -292,7 +285,7 @@ func (k Keeper) calcUserAllocations(
 		return userAllocations, nil
 	}
 
-	tokensPerPoint := za.AmountOf("uqck").ToDec().Quo(sum)
+	tokensPerPoint := zone.ValidatorSelectionAllocation.AmountOfNoDenomValidation("uqck").ToDec().Quo(sum)
```
joe-bowman, line 288: also here :)

joe-bowman (changes requested):
lgtm - but if we can replace the explicit `uqck` references with a `stakingKeeper.BondDenom(ctx)` call, then we'll avoid a footgun later on.

x/participationrewards/keeper/rewards_validatorSelection.go
zanicar: done

PR #120, opened 2022-06-25
gofumpt

faddat:
this is reproducible like:

go install mvdan.cc/gofumpt@latest
gofumpt -w -l .

Basically gofumpt formats the code very nicely, making it easier to work with.


joe-bowman:
Gofumpt was run as part of #146

Issue #123, opened 2022-06-26
Make test


faddat:
the command "make test" is failing.

It is failing because it is expecting the cosmos-sdk default minter module.  I think we might be able to fix it by making a PR that contains the minter module from here, into the forked quicksilver-cosmos-sdk

joe-bowman:
Make test now passes, and is running with every commit :)

PR #124, opened 2022-06-27
test-fix: updated test suite

zanicar:
- ics cli test updated to use correct encoding
- ics test suite updated to reflect implemented cli changes
- valid tests removed as they require IBC for verification

TODO: implement IBC based test for valid cases

Addresses #123


faddat:
thanks so much!

joe-bowman (approved):
lgtm

PR #128, opened 2022-06-30
Cross platform binaries, build cache

shmutalov:
- Added cross-platform binary build: `linux-amd64`, `linux-arm64`, `windows-amd64`, `windows-arm64`
- Added build caching (improves build and test time)
- Added built binaries saving as build artefacts (Github automatically creates zip file for each build platform binary)

### DIFF
--- a/.github/workflows/buildtest.yaml
+++ b/.github/workflows/buildtest.yaml
@@ -10,6 +10,10 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


joe-bowman (approved):
lgtm - not sure we want artifacts build on every run - maybe on merges to main - but we'll see how it goes :)

shmutalov:
It is also configurable, we can save artifacts only when `main` branch receives the changes

PR #130, opened 2022-07-03
participation-rewards bug fixes

zanicar:
- SetValidatorForZone now sets the validator score to zero;
- HandleUpdateZoneProposal now returns the error after logging;
- calcUserValidatorSelectionAllocation ensures a non-nil score;
- Dockerfile bumped Go version to 1.18;


joe-bowman (approved):
lgtm

PR #135, opened 2022-07-11
Handle if LastEmissionHeight is nil

joe-bowman:
- post upgrade, existing queries will have a nil LastEmissionHeight. Handle this case.
- add no-op upgrade handler for underpressure release.


zanicar (approved):
lgtm

PR #141, opened 2022-07-27
x/interchainstaking: keeper, types: fix non-determinism from map iteration

### LINKED ISSUE #140: x/interchainstaking/keeper: (*Keeper).UpdateDelegationRecordsForAddress re-fetching proofs for outdated records happens non-deterministically
If we examine this code in (*Keeper).UpdateDelegationRecordsForAddress https://github.com/ingenuity-build/quicksilver/blob/41136a59e0dc0e1e5d07472e805d5944ef326b3b/x/interchainstaking/keeper/ibc_handlers.go#L547-L565

we can see that we iterate over the map of unique delegator addresses and if at all there is an outdated record, we refetch it. However, that code uses a map which is non-deterministic. Might be useful for us to have a deterministic way of fetching outdated records to have consistent logs, and predictable behavior. The fix for this is to fetch validator addresses from the map, sort them and then later iterate on the sorted slice of unique validator addresses.

This is a low to medium issue but if some ordering issue exists in the record fetches it can be a high severity issue /cc @kirbyquerby @elias-orijtech. I shall send a PR for it.

odeke-em:
In removing stale delegations, the loop was non-deterministic due to
iterating over a map. This change fixes that by extracting the keys
firstly into a slice, sorting it then iterating using those.

Also while here fixed non-determinism in types.Error which produced
different results on each invocation of types.Error.Error() and added
a unit test to lock this behavior in and avoid regressions.

Fixes #140

### DIFF
--- a/x/interchainstaking/keeper/ibc_handlers.go
+++ b/x/interchainstaking/keeper/ibc_handlers.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strings"
 
 	//lint:ignore SA1019 ignore this!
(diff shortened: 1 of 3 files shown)


zanicar (commented):
lgtm

joe-bowman (approved):
lgtm

PR #143, opened 2022-07-27
x/interchainstaking/keeper: fix non-determinism in msgServer.RequestRedemption

### LINKED ISSUE #142: x/interchainstaking/keeper: (msgServer).RequestRedemption redelegation and transaction submission is non-deterministic
If we examine this code https://github.com/ingenuity-build/quicksilver/blob/41136a59e0dc0e1e5d07472e805d5944ef326b3b/x/interchainstaking/keeper/msg_server.go#L145-L155

we can see that we iterate by key on msgs which is a map, and by map iteration in Go it is guaranteed to be randomized. There is an impact on the order of submitted transactions. The fix for this is grabbing out the keys, sorting them and then iterating on the sorted keys.

/cc @kirbyquerby @elias-orijtech. I am sending a PR for this issue.

odeke-em:
Previously the code that invoked keeper.SubmitTx relied on
the range order from map iteration, which in Go is guaranteed
to be randomized hence non-deterministic. This change fixes that
by firstly grabbing out delegator addresses into a slice, sorting
that slice and then iterating on the sorted slice.

Fixes #142

### DIFF
--- a/x/interchainstaking/keeper/msg_server.go
+++ b/x/interchainstaking/keeper/msg_server.go
@@ -6,6 +6,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"sort"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/bech32"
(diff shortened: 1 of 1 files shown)


joe-bowman (approved):
lgtm

PR #145, opened 2022-07-29
make transactions provable

joe-bowman:
- bump cosmos-sdk to v0.45.6-ls.1
- bump ibc-go to v3.1.0
- add relayer Dockerfile and test container
- add `GetTxWithProofResponse` as extensions of Cosmos-SDK and `GetTxResponse` types, to include Tendermint `TxProof` and `Header` fields for validation.
- add proto dependencies for the above


zanicar (approved):
lgtm

PR #147, opened 2022-07-29
Lint fixes and golint-ci as a GH Action

joe-bowman:
Fixes from golint-ci:

Mostly:
 - catching returned err and handling properly
 - convert snake_case to camelCase
 - convert somethingId to somethingID 🤷🏻 
 - remove unused params
 - removed unused return values (errs that can only ever be nil)

Also adds golint-ci on push and pr via Github Actions


zanicar (approved):
lgtm

Issue #154, opened 2022-08-01
x/interchainstaking/keeper: AccountBalanceCallback should avoid a runtime panic by invoking sdk.ValidateDenom instead of crashing


odeke-em:
If we examine the code in https://github.com/ingenuity-build/quicksilver/blob/99b1e83ca9c0ede6f54c152a7895870d2ea9c87c/x/interchainstaking/keeper/callbacks.go#L431
we can see that it doesn't return an error but really inside it is comprised of this code in the cosmos-sdk https://github.com/cosmos/cosmos-sdk/blob/a1143138716b64bc4fa0aa53c0f0fa59eb675bb7/types/coin.go#L16-L27 per
![newCoinPanic](https://user-images.githubusercontent.com/4898263/182076152-3f250fa9-a352-4a26-8318-a0d4c37410af.png)

## Remedy
We can invoke sdk.ValidateDenom and then return an error as the signature of AccountBalanceCallback provides for so
```go
if err := sdk.ValidateDenom(denom); err != nil {
    return fmt.Errorf("failed to validate coin denomination %q, got %w", err)
}
coin = sdk.NewCoin(denom, sdk.ZeroInt()) 
```

joe-bowman:
This should be okay, as the AccountBalanceCallback is called on callback from a provable query - so the query must be validated against the remote chain's state for the callback to be triggered (and thus, the denom must be valid at this point) - but that doesn't protect against future unintentional misuse of the callback, so will add this for protection from that. Thank you :)

joe-bowman:
closed by #155

Issue #153, opened 2022-08-01
x/interchainstaking/types: simplify and optimize indentString to use strings.Repeat instead of manual string concatenation n times


odeke-em:
If we look at this code https://github.com/ingenuity-build/quicksilver/blob/99b1e83ca9c0ede6f54c152a7895870d2ea9c87c/x/interchainstaking/types/error.go#L57-L64 , it essentially goes through d times and manually builds a string via concatenation of the indent string but really it can be simplified to just (and improve readability by changing d to n)
```go
 func indentString(indent string, n int) string { 
     return strings.Repeat(indent, n)
}
```

joe-bowman:
closed by #155

Issue #152, opened 2022-08-01
x/interchainstaking/keeper: utilize known map size to generate keys slice in (ValidatorIntents).Keys


odeke-em:
If we look at this code in https://github.com/ingenuity-build/quicksilver/blob/99b1e83ca9c0ede6f54c152a7895870d2ea9c87c/x/interchainstaking/types/delegation.go#L112-L122
we can improve its performance by instead doing the following

```go
 keys := make([]string, 0, len(v)) 
 for key := range v { 
 	keys = append(keys, key) 
 } 
 sort.Strings(keys) 
  
 return keys
```

or even faster

```go
 keys := make([]string, len(v)) 
 i := 0
 for key := range v { 
 	keys[i] = key
        i++
 } 
 sort.Strings(keys) 
  
 return keys
```

and here are some performance studies showing benefits per https://bencher.orijtech.com/perfclinic/sliceupdate and in particular the shoot-out results in https://bencher.orijtech.com/perfclinic/sliceupdate/#results

joe-bowman:
closed by #155

PR #155, opened 2022-08-01
fixes and improvements

zanicar:
- ValidatorIntents map created using known size;
 - errors strings indent now uses strings std lib;
 - AccountBalanceCallback now returns error on invalid denom;

Addresses #152, #153 and #154


joe-bowman (approved):
lgtm

PR #157, opened 2022-08-01
Ics import export

joe-bowman:
**State Breaking change**

This change add Genesis import/export logic to x/interchainstaking


zanicar (approved):
lgtm

PR #159, opened 2022-08-02
airdrop: refactor wip

zanicar:
- added airdrop Status enum;
 - implemented Msg service for Claim;
 - implemented Proposal to register zonedrops;
 - added AccountBalance query;
 - implemented cli queries;
 - implemented cli transactions;
 - implemented invariants;
 - implemented multi-error for module...
   * this should probably be externalized for reuse;


proto/quicksilver/airdrop/v1/messages.proto
```diff
@@ -8,4 +8,26 @@ import "google/api/annotations.proto";
...
+}
+
+message MsgClaim {
+    option (gogoproto.equal) = false;
+    option (gogoproto.goproto_getters) = false;
+
+    string chain_id = 1 [ (gogoproto.moretags) = "yaml:\"chain_id\"" ];
+    int32 action = 2 [ (gogoproto.moretags) = "yaml:\"action\"" ];
+    string address = 3 [ (gogoproto.moretags) = "yaml:\"address\"" ];
+}
```
joe-bowman, line 27: This needs to be able to take some arbitrary []bytes because the payload that needs to be verified will differ between action types (or nil for local-chain claims).

joe-bowman: For example; the osmosis lockup proofs look like this:

https://gist.github.com/joe-bowman/e1eff2fcdf68d6ef5b9ffcf89a04d7d8

So we can either make the payload one of OsmosisPayload|CrescentPayload|... or free []bytes that we validate later.

This latter suggestion will work better when it comes to wasm based validations, as we won't _want_ to define data structures in the module then.

proto/quicksilver/airdrop/v1/proposals.proto
```diff
@@ -0,0 +1,19 @@
...
+message RegisterZoneDropProposal {
+    option (gogoproto.equal) = false;
+    option (gogoproto.goproto_getters) = false;
+    option (gogoproto.goproto_stringer) = false;
+
+    string title = 1;
+    string description = 2;
+
+    ZoneDrop zone_drop = 3 [ (gogoproto.moretags) = "yaml:\"zone_drop\"" ];
+    repeated ClaimRecord claim_records = 4 [ (gogoproto.moretags) = "yaml:\"claim_records\"" ];
```
joe-bowman, line 18: If we leave this as is, it means all the claim records must be submitted as json (making the payload far larger than it needs to be). If claim_records is a byte array, we can compress / optimise this payload to our heart's content, and expand this once the tx is in the application - thoughts?

joe-bowman: It is possible that the unpacking might be more costly in gas that the huge tx. But we ought to try this out.

x/airdrop/types/errors.go
```diff
@@ -1,11 +1,63 @@
...
 
 // x/airdrop module sentinel errors
 var (
 	ErrZoneDropNotFound    = sdkerrors.Register(ModuleName, 1, "zone airdrop not found")
 	ErrClaimRecordNotFound = sdkerrors.Register(ModuleName, 2, "claim record not found")
+	ErrUnknownStatus       = sdkerrors.Register(ModuleName, 3, "unknown status")
 )
+
+// NewMultiError returns an error aggregate using the given map.
+func NewMultiError(errors map[string]error) Errors {
```
joe-bowman, line 19: iirc, we have this in ICS already? Do we want to extract this out into the global `utils` package?

x/airdrop/types/events.go
```diff
@@ -1,5 +1,8 @@
 package types
 
 const (
-	EventTypeClaim = "airdrop claim"
+	EventTypeClaim            = "airdrop claim"
```
joe-bowman, line 4: do events need to be underscore rather than space delimited? I'm not sure where these are used?

joe-bowman (changes requested):
overall looks good; some points need clarifying.

proto/quicksilver/airdrop/v1/messages.proto
zanicar: I favor the []byte route... more flexible.

proto/quicksilver/airdrop/v1/proposals.proto
zanicar: I agree, we should have the proposal payload as []byte (containing both ZoneDrop & []ClaimRecord); This will correspond with RegisterZoneProposals which are once-off type transactions, decompression overhead feels like the better option here vs massive uncompressed json blob.

x/airdrop/types/errors.go
zanicar: Yes, I have it as a note in the commit message 204f417...
Though it would be best as a stand-alone micro lib imho

x/airdrop/types/events.go
zanicar: yes indeed it should be underscore...

x/airdrop/types/proposals.go
```diff
@@ -0,0 +1,48 @@
...
+func (m RegisterZoneDropProposal) String() string {
+	var b strings.Builder
+
+	b.WriteString("Airdrop - ZoneDrop Registration Proposal:\n")
+	b.WriteString(fmt.Sprintf("\tTitle:       %s\n", m.Title))
+	b.WriteString(fmt.Sprintf("\tDescription: %s\n", m.Description))
+	b.WriteString("\tZoneDrop:\n")
+	b.WriteString(fmt.Sprintf("\n%v\n\n", m.ZoneDrop))
+	b.WriteString("\tClaimRecords:\n")
+	b.WriteString(fmt.Sprintf("\n%v\n", m.ClaimRecords))
```
joe-bowman, line 45: Not sure how useful it'll be in practice to output all the claim records data in this string?

joe-bowman (approved):
lgtm

x/airdrop/types/proposals.go
zanicar: only useful for debugging and early testing... thereafter it can be dropped as I'm sure nobody really wants to read a raw byte slice

joe-bowman (approved):
lgtm

joe-bowman (approved):
lgtm

Issue #151, opened 2022-07-30
interchainstaking/types/Receipt should not store a copy of Zone


joe-bowman:
This could well explain the memory leak and storage bloat in Rhapsody/Killerqueen, as every handled transaction receipt stores an entire copy of the zone. Recent versions have reduced the size of this struct, but storing many instances of this is clearly suboptimal!

PR #161, opened 2022-08-03
#151 Do not store a copy of Zone in DepositTx Receipt

joe-bowman:
STATE BREAKING

Do not store a copy of zone in Receipt struct; this leads to state and memory bloat.


zanicar (approved):
lgtm

PR #165, opened 2022-08-03
162 fetch height at boundary

joe-bowman:
merged #164 into the wrong branch 🤦🏻 

- fetch height at boundary via ICQ callback
- store boundary_height against zone to be used by PR
- changelog
- formatting


zanicar (approved):
lgtm

Issue #170, opened 2022-08-08
Defer epochly distribution for epochs 1+2


joe-bowman:
In order that users have ample opportunity to bond prior to QCK staking rewards disbursement, we are going to defer staking rewards for the first two epochs.

This should be a reasonably small change that stops tokens being emitted when currentEpoch < 3.

PR #171, opened 2022-08-08
participation-rewards: defer distribution

zanicar:
- defer initial rewards distribution until 3rd epoch;
 - solves #170;


joe-bowman (approved):
lgtm

PR #176, opened 2022-08-15
interchainstaking: now uses internal multierror

### LINKED ISSUE #175: x/interchainstaking: remove custom MultiError and use global lib
We now have a copy of MultiError in x/interchainstaking and internal; we need to remove from the former.

There are references in:
types/error.go
types/msgs.go
keeper/msg_server.go

zanicar:
closes #175


joe-bowman (approved):
lgtm.

PR #177, opened 2022-08-15
Airdrop module

zanicar:


x/airdrop/keeper/claim_action.go
```diff
@@ -0,0 +1,138 @@
...
+	if !ok {
+		return fmt.Errorf("zone not found for %s", cr.ChainId)
+	}
+
+	dAmount := sdk.NewCoin(
+		zone.BaseDenom,
+		threshold.MulInt64(int64(cr.BaseValue)).TruncateInt(),
+	)
+
+	if !k.bankKeeper.HasBalance(ctx, addr, dAmount) {
```
joe-bowman, line 72: We want this to succeed even if users have sent their tokens to Osmosis, for example.

This needs to be a lookup against receipts for that (user|zone) tuple.

We added an iterator on receipts for user/zone, so we should trivially be able to add a function to fetch the summed value of deposits for a given zone.

x/airdrop/types/airdrop.go
```diff
@@ -2,6 +2,7 @@ package types
 
 import (
 	"fmt"
+	time "time"
```
joe-bowman, line 5: surprised this go through linting :/

x/airdrop/types/airdrop.go
```diff
@@ -70,11 +79,21 @@ func (cr ClaimRecord) ValidateBasic() error {
...
+			kstr := fmt.Sprintf("ActionsCompleted[%d]", i)
 			if int(ae) >= len(Action_name) {
-				kstr := fmt.Sprintf("ActionsCompleted[%d]", i)
-				errors[kstr] = fmt.Errorf("enum out of bounds, expects [0-%d), got %d", len(Action_name), ae)
+				errors[kstr+" Enum"] = fmt.Errorf("%w, got %d", ErrActionOutOfBounds, ae)
 			}
+			// calc sum
 			sum += ca.ClaimAmount
+			// check completed time
+			if ca.CompleteTime.After(time.Now()) {
```
joe-bowman, line 90: this could be non-deterministic. time.Now() will differ between nodes - use ctx.BlockTime() instead.

x/airdrop/types/airdrop.go
zanicar: bad linter... auto import does that way too often for my liking...

x/airdrop/types/airdrop.go
zanicar: this is in ValidateBasic, which doesn't have access to ctx...
for the purposes here we only need to ensure that the CompleteTime is in the past.
Any recommendations?

zanicar: we could move this check from stateless to stateful validation...

x/airdrop/keeper/claim_action.go
zanicar: are we looking for the net deposit amount (deposits - redemptions) or simply gross deposits?

x/airdrop/types/airdrop.go
joe-bowman: This is only for invariant checking? If so, we can probably let it slide as long as we comment in the code here the risk that _significant_ clock drift may cause nodes to fail invariant checking.

x/airdrop/keeper/claim_action.go
joe-bowman: net deposit is fine

joe-bowman (approved):
lgtm.

PR #179, opened 2022-08-16
Release v0.6.1

joe-bowman:
- Implement unbonding
- bump osmosis test container
- update to icq query to support go relayer


zanicar (commented):
so far so good...
just a couple of minor things for clarification maybe

zanicar (approved):
lgtm

PR #190, opened 2022-08-26
remediation: updated panics and error handling

joe-bowman:


x/airdrop/keeper/abci.go
```diff
@@ -12,6 +12,7 @@ func (k Keeper) BeginBlocker(ctx sdk.Context) {
 func (k Keeper) EndBlocker(ctx sdk.Context) {
 	for _, zd := range k.UnconcludedAirdrops(ctx) {
 		if err := k.EndZoneDrop(ctx, zd.ChainId); err != nil {
+			// failure in EndBlocker should panic
 			panic(err)
```
joe-bowman, line 16: I thought we explicitly _didn't_ panic in Blockers? It will crash the chain.

zanicar: ooo, an unfix... let me "fix" it

zanicar (approved):
looking in the mirror...

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -830,7 +830,12 @@ func (k *Keeper) prepareRewardsDistributionMsgs(zone types.Zone, rewards sdk.Coi
 	var msgs []sdk.Msg
 
 	dust := rewards.Amount
-	portion := rewards.Amount.ToDec().Quo(sdk.NewDec(int64(len(zone.DelegationAddresses)))).TruncateInt()
+	numDelegators := len(zone.DelegationAddresses)
+	if numDelegators == 0 {
+		// zone has no delegators, we must panic here, something is very wrong...
+		panic("internal zone error, zone has no delegators")
```
joe-bowman, line 836: @zanicar - I think this can plausibly happen if the zone has only just been registered and an epoch hits. We should throw and error here, and catch it when iterating the zones, and just continue to the next.

zanicar:
@joe on the surface this looks like a potential edge case error, but I do believe it to be a legitimate panic. Logically we must have at least one delegator to reach this code:

- this is in `prepareRewardsDistributionMsgs`
- which in turn is called from `DistributeRewardsFromWithdrawAccount`
- which is a callback triggered from within `HandleWithdrawRewards`
- which is called from `HandleAcknowledgement` in response to a `MsgWithdrawDelegatorReward`
- which is submitted from `WithdrawDelegationRewardsForResponse`
- which is created in an `IterateDelegatorDelegations`
- which uses a delegator address obtained from `QueryDelegationTotalRewardsResponse`
- which implies at least one delegator must exist at this point

https://github.com/ingenuity-build/quicksilver/blob/20dc658480b1af1cb323b4ab4a8e5925ee79a0ed/x/interchainstaking/keeper/delegation.go#L216-L224

unless I am missing something?

joe-bowman:
Nope, this works for me - let's close #238 :)

Issue #183, opened 2022-08-25
x/interchainstaking/types: missing range check for weight in ConvertMemoToOrdinalIntents


elias-orijtech:
A comment in ConvertMemoToOrdinalIntents states that the memo weight is in the range 1-200:

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/types/zones.go#L104

However, there is no corresponding check in the code parsing the weight:

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/types/zones.go#L110-L111

CC @odeke-em

Issue #181, opened 2022-08-25
x/interchainstaking/keeper: suspicious code in parseDelegationKey


elias-orijtech:
The `parseDelegationKey` functions contains a commented out variant of its parsing:

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/keeper/ibc_packet_handlers.go#L584-L593

The `valAddrLen` should be removed altogether, or the parsing implementation updated to match reality. In either case, a test should verify that parseDelgationKey really does match the expected format of a delegation key, because it is only called from a special (rare?) case in a callback:

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/keeper/callbacks.go#L167-L169

CC @odeke-em

joe-bowman:
`valAddrLen` field was originally ignored, because we can simply read to the end of the slice. But I have added validation and out-of-bounds checking to the method - and caught another bug where I'd used python style [start:length] instead of [start:end] syntax.

Good catch!

PR #192, opened 2022-08-26
remediations from orijtech audit

### LINKED ISSUE #180: x/airdrop: redundant code in InitGenesis by placing zero value into map[string]uint64:: low priority
This code https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/airdrop/genesis.go#L20-L22 is redundant and unnecessary given that in Go, the zero value of primitive types is useful so that code can simply be deleted and
https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/airdrop/genesis.go#L23 will simply work and accumulate with the zero value and this is an exhibit https://go.dev/play/p/3aXlcTHIuIn

```go
package main

func main() {
	// Simplified version.
	m := make(map[string]uint64)
	m["e"] += 2
	println(m["e"])

	// Redundant version.
	n := make(map[string]uint64)
	if _, ok := n["e"]; !ok {
		n["e"] = 0
	}
	n["e"] += 2
	println(n["e"])
}
```

which produces
```shell
2
2
```

### LINKED ISSUE #182: x/interchainstaking/types: inconsistent error handling in ConvertMemoToOrdinalIntents
Missing error check when converting memo to valoper address:

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/types/zones.go#L115

Perhaps addresses are always valid by higher level logic, but then earlier error checks are redundant:

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/types/zones.go#L98-L107

CC @odeke-em

### LINKED ISSUE #184: x/interchainstaking/keeper: possible missing return in Keeper.BeginBlocker error path
In `Keeper.BeingBlocker`

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/keeper/abci.go#L23

there is a panic commented out and a comment explaining why. However, there is no `return` statement or similar to halt further execution of the method. If that is intentional, I suggest a comment explaining why continuing is OK.

Same comment applies to the next error path,

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/keeper/abci.go#L25-L27

where a missing return may be OK because it completes the `blockheight%10` block.

CC @odeke-em

### LINKED ISSUE #187: x/interchainstaking/keeper: possible missing out-of-bounds check in AccountBalanceCallback
There's no length check before slicing:

https://github.com/ingenuity-build/quicksilver/blob/a7f142ab99e978482f66248d0009b90820d0f0ce/x/interchainstaking/keeper/callbacks.go#L432

There may be a higher-level reason that `len(request.Query)` is always greater than zero.

CC @odeke-em

joe-bowman:
- add comments explaining the non-return under error conditions in x/interchainstaking beginBlocker; fixes #184
- remove redundant code; fixes #180
- refactor, test and bug fix ParseDelegationKey #181
- improved error handling; closes #182 #183
- improved input validation; closes #187

### DIFF
--- a/x/interchainstaking/keeper/msg_server.go
+++ b/x/interchainstaking/keeper/msg_server.go
@@ -112,7 +112,10 @@ func (k msgServer) RequestRedemption(goCtx context.Context, msg *types.MsgReques
 
 	intentMap := userIntent.ToAllocations(nativeTokens)
 
-	targets := k.GetRedemptionTargets(ctx, *zone, intentMap) // map[string][string]sdk.Coin
+	targets, err := k.GetRedemptionTargets(ctx, *zone, intentMap) // map[string][string]sdk.Coin
+	if err != nil {
+		return nil, err
+	}
 
 	if len(targets) == 0 {
 		return nil, fmt.Errorf("targets can never be zero length")
@@ -151,7 +154,7 @@ func (k msgServer) RequestRedemption(goCtx context.Context, msg *types.MsgReques
 			}
 			sumAmount = sumAmount.Add(target.Value[0])
 			if _, found := k.GetWithdrawalRecord(ctx, zone, hashString, target.DelegatorAddress, target.ValidatorAddress); found {
-				fmt.Printf("cannot withdraw twice for the same delegator/validator tuple in a single transaction")
+				return nil, fmt.Errorf("cannot withdraw twice for the same delegator/validator tuple in a single transaction")
 			}
 			k.Logger(ctx).Info("Store", "del", target.DelegatorAddress, "val", target.ValidatorAddress, "hash", hashString, "chain", zone.ChainId)
 			k.AddWithdrawalRecord(ctx, zone, target.DelegatorAddress, target.ValidatorAddress, msg.DestinationAddress, target.Value[0], msg.Value, hashString, time.Unix(0, 0))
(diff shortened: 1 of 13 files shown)


x/interchainstaking/keeper/msg_server.go
```diff
@@ -169,7 +172,7 @@ func (k msgServer) RequestRedemption(goCtx context.Context, msg *types.MsgReques
 	for _, delegator := range delegators {
 		icaAccount, err := zone.GetDelegationAccountByAddress(delegator)
 		if err != nil {
-			panic(err) // panic here because something is terribly wrong if we can't find the delegation bucket here!!!
+			return nil, err // panic here because something is terribly wrong if we can't find the delegation bucket here!!!
```
zanicar, line 175: maybe we should keep the panic here?

zanicar (commented):
just the one panic we should maybe keep...

x/interchainstaking/keeper/msg_server.go
joe-bowman: Done

zanicar (approved):
lgtm

PR #206, opened 2022-08-27
Further audit fixes

### LINKED ISSUE #201: x/interchainstaking/keeper: error result from GetValidatorByValoper used as found bool
This code uses the error result from `Zone.GetValidatorByValoper` as a found sentinel:

https://github.com/ingenuity-build/quicksilver/blob/8c558f01b2b2528bd36c484da2c0b5d64e2b110b/x/interchainstaking/keeper/keeper.go#L194-L223

That's ok because a missing validator is the only kind of error. However, to aid in clarity I suggest changing `GetValidatorByValoper` to explicitly return a bool to make it clear no other kinds of errors are possible.

CC @odeke-em

joe-bowman:
- make GetValidatorByValoper() return a found sentinel, instead of an err; fixes #201

### DIFF
--- a/x/interchainquery/types/codec.go
+++ b/x/interchainquery/types/codec.go
@@ -14,7 +14,7 @@ var (
 )
 
 func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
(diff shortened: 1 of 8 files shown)


zanicar (approved):
lgtm

Issue #220, opened 2022-08-29
x/airdrop/keeper: Keeper.decompress is too complex


elias-orijtech:
https://github.com/ingenuity-build/quicksilver/blob/4ea3aa77d56b674c7189835064fda16f6bebf8da/x/airdrop/keeper/proposal_handler.go#L74-L100

can be replaced with (untested):

```
func (k Keeper) decompress(data []byte) ([]byte, error) {
	zr, err := zlib.NewReader(bytes.NewReader(data))
	if err != nil {
		return nil, err
	}
	defer zr.Close()
	return io.ReadAll(zr)
}
```

CC @odeke-em

zanicar:
No, this introduces a DoS vulnerability via a decompression bomb...

elias-orijtech:
It's true that the replacement is vulnerable to decompression bombs. However, so was the original because there is no limit to the amount of decompressed data. In fact, `io.ReadAll` contains code similar to `Keeper.decompress`: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/io/io.go;l=654.

odeke-em:
Yeah, @zanicar as @elias-orijtech just pointed out, I too don't see how the original code prevents a zip bomb except by using bytes.Buffer that grows incrementally versus slice reallocations in 1 shot. The first link you provided ensures that if for example optimizations like sendfile are applied that io.Copy can take advantage of them, but in the instance of the code in this repo, it doesn't.

joe-bowman:
This has been closed in earlier audit remediation fixes.

Issue #219, opened 2022-08-29
x/airdrop/keeper: inconsistent claim record validation


elias-orijtech:
The claim record validations at

https://github.com/ingenuity-build/quicksilver/blob/4ea3aa77d56b674c7189835064fda16f6bebf8da/x/airdrop/keeper/claim_record.go#L131-L133

and

https://github.com/ingenuity-build/quicksilver/blob/4ea3aa77d56b674c7189835064fda16f6bebf8da/x/airdrop/keeper/claim_record.go#L176-L178

seems insufficient: why is check for the empty address enough? If non-empty, yet invalid, addresses are ok, why have the check at all?

CC @odeke-em

joe-bowman:
Fixed in #221

Issue #218, opened 2022-08-29
x/interchainstaking/types: brittle use of NewDecFromStr


elias-orijtech:
The initialization of `DefaultCommissionRate`,

https://github.com/ingenuity-build/quicksilver/blob/4ea3aa77d56b674c7189835064fda16f6bebf8da/x/interchainstaking/types/params.go#L17

is needlessly brittle. I suggest using the error-free `sdk.NewDec` constructor, or at least the error-checking `sdk.MustNewDecFromStr` variant.

CC @odeke-em

joe-bowman:
Closed by #221

Issue #217, opened 2022-08-29
x/interchainstaking/types: suspicious TODO in MsgRequestRedemption.ValidateBasic


elias-orijtech:
A TODO in a validate function is often (but not always) a sign of missing validation:

https://github.com/ingenuity-build/quicksilver/blob/4ea3aa77d56b674c7189835064fda16f6bebf8da/x/interchainstaking/types/msgs.go#L33

CC @odeke-em

elias-orijtech:
Another in `MsgSignalIntent.ValidateBasic`:

https://github.com/ingenuity-build/quicksilver/blob/4ea3aa77d56b674c7189835064fda16f6bebf8da/x/interchainstaking/types/msgs.go#L113

joe-bowman:
Resolved in #221

PR #221, opened 2022-08-30
Remediation: Audit fixes and updates

zanicar:
Resolves the following:

- #185 
- #195 
- #197 
- #199 
- #203 
- #209 
- #210 
- #211 
- #216 
- #217 
- #218 
- #219 
- #220

### DIFF
--- a/x/airdrop/keeper/claim_record.go
+++ b/x/airdrop/keeper/claim_record.go
@@ -128,9 +128,6 @@ func (k Keeper) GetClaimableAmountForAction(ctx sdk.Context, chainID string, add
 	if err != nil {
 		return 0, err
 	}
-	if cr.Address == "" {
-		return 0, nil
-	}
 
 	// action already completed, nothing to claim
 	if _, exists := cr.ActionsCompleted[int32(action)]; exists {
@@ -138,7 +135,7 @@ func (k Keeper) GetClaimableAmountForAction(ctx sdk.Context, chainID string, add
 	}
 
 	// get zone airdrop details
-	zd, ok := k.GetZoneDrop(ctx, chainID)
+	zd, ok := k.GetZoneDrop(ctx, cr.ChainId)
 	if !ok {
 		return 0, types.ErrZoneDropNotFound
 	}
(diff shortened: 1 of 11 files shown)


x/interchainstaking/types/msgs.go
```diff
@@ -57,23 +63,28 @@ func (msg MsgRequestRedemption) GetSigners() []sdk.AccAddress {
...
 // The combined weights must be 1.0 and the valoper addresses must be valid
-// bech32 strings. (what about zero weights?)
+// bech32 strings.
 //
 // Tokens are comma separated, e.g.
 // "0.3cosmosvaloper1xxxxxxxxx,0.3cosmosvaloper1yyyyyyyyy,0.4cosmosvaloper1zzzzzzzzz".
 func IntentsFromString(input string) ([]*ValidatorIntent, error) {
+	// regexp defining what an intent looks like:
+	// {value}{address}
 	iexpr := regexp.MustCompile(`(\d.\d+)(.+1\w+)`)
```
joe-bowman, line 77: The `.` should be escaped here, else it'll match any character - e.g. this would validate `0L3cosmosvaloper1xxxxxx`

x/participationrewards/types/msgs.go
```diff
@@ -17,22 +18,43 @@ func NewMsgSubmitClaim(
...
-// ValidateBasic Implements Msg.
+// ValidateBasic implements Msg: stateless checks.
 func (msg MsgSubmitClaim) ValidateBasic() error {
 	errors := make(map[string]error)
 	if _, err := sdk.AccAddressFromBech32(msg.UserAddress); err != nil {
 		errors["UserAddress"] = err
 	}
 
-	// TODO: check for valid zone (chainID)
+	if msg.Zone == "" {
```
joe-bowman, line 34: not a problem, just wondered if there was a reason for `== ""` here rather than using `len()`

x/airdrop/keeper/claim_record.go
```diff
@@ -128,17 +128,14 @@ func (k Keeper) GetClaimableAmountForAction(ctx sdk.Context, chainID string, add
 	if err != nil {
 		return 0, err
 	}
-	if cr.Address == "" {
```
joe-bowman, line 131: Do we validate the address prior to this?

x/interchainstaking/types/msgs.go
zanicar: good catch

x/participationrewards/types/msgs.go
zanicar: yes len() is a better approach

x/airdrop/keeper/claim_record.go
zanicar: yes, this is after `GetClaimRecord` that calls `AccAddressFromBech32`...
thus, an empty address is not possible at this point (previously needed as our initial implementation was based on Stargaze claim module which allowed an empty address at this point).

joe-bowman (approved):
lgtm

PR #222, opened 2022-08-30
Release/v0.6.3

### LINKED ISSUE #194: x/interchainstaking/keeper: convoluted code in AccountBalanceCallback
The for loop in

https://github.com/ingenuity-build/quicksilver/blob/00a76ee447b614471ef5aded5c70fd88c9fce70a/x/interchainstaking/keeper/callbacks.go#L446-L457

can be replaced with a call to `bytes.Index`, shortening the code and avoid abusing `sdk.ValidateDenom` to reject a missing denomination. Something like (untested):

```
		idx := bytes.Index(query.Request, accAddr)
		if idx == -1 {
			return errors.New("AccountBalanceCallback: invalid request query")
		}
		denom := string(query.Request[idx:])
```

CC @odeke-em

joe-bowman:
- add upgrade handler for innuendo1 upgrade
- add tests for upgrade handler
- fix #194 simplify AccountBalanceCallback logic

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -1,0 +1,46 @@
+package app
+
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	airdroptypes "github.com/ingenuity-build/quicksilver/x/airdrop/types"
+)
+
+func GetInnuendo1Upgrade(app *Quicksilver) types.UpgradeHandler {
+
+	return func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+		ReplaceZoneDropChain(ctx, app, "osmotest-4", "osmo-test-4", ctx.BlockHeader().Time)
+		return app.mm.RunMigrations(ctx, app.configurator, fromVM)
+	}
+}
+
+// replaces zonedrop and claimrecords for a given chain, with another chain and update start time.
+func ReplaceZoneDropChain(ctx sdk.Context, app *Quicksilver, chainIdFrom string, chainIdTo string, start time.Time) {
+	ad, found := app.AirdropKeeper.GetZoneDrop(ctx, chainIdFrom)
+	if !found {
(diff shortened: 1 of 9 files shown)


app/upgrades.go
```diff
@@ -0,0 +1,71 @@
...
+	); err != nil {
+		panic(err)
+	}
+
+	app.AirdropKeeper.DeleteZoneDrop(ctx, chainIdFrom)
+
+	// update unbonding time to 48h.
+	stakeParams := app.StakingKeeper.GetParams(ctx)
+	stakeParams.UnbondingTime = 48 * time.Hour
+	app.StakingKeeper.SetParams(ctx, stakeParams)
```
odeke-em, line 70: Is this specified in the specification and does it have any relation to the Quicksilver Epoch?

x/interchainstaking/keeper/callbacks.go
```diff
@@ -451,18 +452,14 @@ func AccountBalanceCallback(k Keeper, ctx sdk.Context, args []byte, query icqtyp
...
-		// if balance is nil, the response sent back is nil, so we don't receive the denom. Override that now.
-		if err := sdk.ValidateDenom(denom); err != nil {
-			return err
+		// if the balance returned is zero for a given denom, we just get a nil response.
+		// lookup the denom from the request so we can set a zero value coin for the correct denom.
+		idx := bytes.Index(query.Request, accAddr)
+		if idx == -1 {
+			return errors.New("AccountBalanceCallback: invalid request query")
 		}
+		denom := string(query.Request[idx:])
```
odeke-em, line 461: How come we are missing the sdk.ValidateDenom check?

app/upgrades.go
joe-bowman: This is only related to the current testnet - the unbonding period was set low to aid unbonding testing on the testnet, but ended up expiring when hermes went down.

In production it will be 21 days.

joe-bowman: I have pushed an update to clarify this change.

x/interchainstaking/keeper/callbacks.go
joe-bowman: My reading of #194 was that the `ValidateDenom` was considered superfluous here. Thoughts?

zanicar (approved):
lgtm

x/interchainstaking/keeper/callbacks.go
joe-bowman: I had re-added ValidateDenom

zanicar (approved):
lgtm

PR #225, opened 2022-08-30
remediation: bug fix

### LINKED ISSUE #224: Bug: Div by zero fix doesn't cater for zero-length slice
Panic thrown in attempt to avoid div-by-zero:
```
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: 10:29PM INF handling epoch end module=x/interchainstaking
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: 10:29PM INF MakeRequest callback=epochblock chain_id=fauxgaia-1 connection_id=connection-0 module=interchainstaking period=-1 query_type=cosmos.base.tendermint.v1beta1.Service/GetLatestBlock request= ttl=0
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: 10:29PM INF taking a snapshot of intents module=x/interchainstaking
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: panic: ordinalized intent sum is zero, this should never happen
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: goroutine 1 [running]:
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: github.com/ingenuity-build/quicksilver/x/interchainstaking/keeper.(*Keeper).AggregateIntents(_, {{0x26d8770, 0xc000052090}, {0x26e5980, 0xc0066b2700}, {{0xb, 0x0}, {0xc0010c9ca0, 0xa}, 0x217f0, ...}, ...}, ...)
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]:         github.com/ingenuity-build/quicksilver/x/interchainstaking/keeper/intent.go:130 +0x57b
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: github.com/ingenuity-build/quicksilver/x/interchainstaking/keeper.Keeper.AfterEpochEnd.func1(_, {{0xc001b065d0, 0xc}, {0xc001b065e0, 0xa}, 0xc001a49180, 0xc001a491d0, 0xc001a49220, {0xc0018a3000, 0x64, ...}, ...})
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]:         github.com/ingenuity-build/quicksilver/x/interchainstaking/keeper/hooks.go:37 +0x2e5
Aug 30 22:29:23 val01-qck-t quicksilverd[442584]: github.com/ingenuity-build/quicksilver/x/interchainstaking/keeper.Keeper.IterateZones({{0x26e72f8, 0xc0001513c0}, {0x26c31f0, 0xc00127e750}, {{0x26e4128, 0xc0001513c0}, {0x26c31f0, 0xc00127e6d0}, {0x26c3218, 0xc00127e7c0}, ...}, ...}, ...)
```

In #192 we add a panic to `x/interchainstaking/keeper/intents.go:aggregateIntents()` to avoid the potential div by zero; however, we did not cater for the case where by `intents` in a zero-length slice (no intents have ever been set) and thus the sum can legitimately be zero without triggering the div-by-zero error.

zanicar:
- now panics when slice is not empty and sum is zero;
- removed relayer from test script;

fixes #224

### DIFF
--- a/x/airdrop/keeper/proposal_handler.go
+++ b/x/airdrop/keeper/proposal_handler.go
@@ -49,9 +49,11 @@ func HandleRegisterZoneDropProposal(ctx sdk.Context, k Keeper, p *types.Register
 			return fmt.Errorf("claim record %d, %w", i, err)
 		}
 
-		for j, ca := range cr.ActionsCompleted {
-			if ctx.BlockTime().Before(ca.CompleteTime) {
-				return fmt.Errorf("invalid zonedrop proposal claim record [%d]: completed action [%d] is in future", i, j)
+		if cr.ActionsCompleted != nil {
+			for j, ca := range cr.ActionsCompleted {
+				if ctx.BlockTime().Before(ca.CompleteTime) {
+					return fmt.Errorf("invalid zonedrop proposal claim record [%d]: completed action [%d] is in future", i, j)
+				}
 			}
 		}
 
--- a/x/airdrop/keeper/proposal_handler.go
+++ b/x/airdrop/keeper/proposal_handler.go
@@ -49,12 +49,8 @@ func HandleRegisterZoneDropProposal(ctx sdk.Context, k Keeper, p *types.Register
 			return fmt.Errorf("claim record %d, %w", i, err)
 		}
 
-		if cr.ActionsCompleted != nil {
-			for j, ca := range cr.ActionsCompleted {
-				if ctx.BlockTime().Before(ca.CompleteTime) {
-					return fmt.Errorf("invalid zonedrop proposal claim record [%d]: completed action [%d] is in future", i, j)
-				}
-			}
+		if len(cr.ActionsCompleted) != 0 {
+			return fmt.Errorf("invalid zonedrop proposal claim record [%d]: contains completed actions", i)
 		}
 
 		if err := k.SetClaimRecord(ctx, cr); err != nil {
--- a/x/airdrop/keeper/proposal_handler.go
+++ b/x/airdrop/keeper/proposal_handler.go
@@ -49,6 +49,12 @@ func HandleRegisterZoneDropProposal(ctx sdk.Context, k Keeper, p *types.Register
 			return fmt.Errorf("claim record %d, %w", i, err)
 		}
 
+		for j, ca := range cr.ActionsCompleted {
+			if ctx.BlockTime().Before(ca.CompleteTime) {
+				return fmt.Errorf("invalid zonedrop proposal claim record [%d]: completed action [%d] is in future", i, j)
+			}
+		}
+
 		if err := k.SetClaimRecord(ctx, cr); err != nil {
 			return fmt.Errorf("invalid zonedrop proposal claim record [%d]: %w", i, err)
 		}
--- a/x/participationrewards/keeper/distribution.go
+++ b/x/participationrewards/keeper/distribution.go
@@ -40,9 +40,12 @@ func (k Keeper) getRewardsAllocations(ctx sdk.Context) rewardsAllocation {
 		k.Logger(ctx).Info("nothing to distribute...")
 
 		// create snapshot of current intents for next epoch boundary
+		// requires intents to be set, no intents no snapshot...
+		// we need to ensure the snapshot is only taken once;
 		// TODO: this needs to be verified as it currently does not trigger anymore
 		for _, zone := range k.icsKeeper.AllZones(ctx) {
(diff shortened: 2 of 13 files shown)


x/participationrewards/keeper/distribution.go
```diff
@@ -40,9 +40,12 @@ func (k Keeper) getRewardsAllocations(ctx sdk.Context) rewardsAllocation {
 		k.Logger(ctx).Info("nothing to distribute...")
 
 		// create snapshot of current intents for next epoch boundary
+		// requires intents to be set, no intents no snapshot...
+		// we need to ensure the snapshot is only taken once;
 		// TODO: this needs to be verified as it currently does not trigger anymore
 		for _, zone := range k.icsKeeper.AllZones(ctx) {
 			for _, di := range k.icsKeeper.AllOrdinalizedIntents(ctx, zone, false) {
+				k.Logger(ctx).Info("TRIGGERED")
```
joe-bowman, line 48: left over debug logging?

zanicar: we have to ensure that intent snapshot is only taken once and not twice...

joe-bowman: this log will trigger once for every user, for ever zone - I suspect it'll be too noisy to be useful?

zanicar: Ok, it seems this is the bootstrapping of intent snapshots...
When we have no intents set the snapshot will not trigger, however if we have intents and a zero module balance we want to take a snapshot of intents to utilize at the next epoch boundary.

So this will not trigger again unless the module balance remains zero (under what reasonable circumstances could this be possible?)

zanicar:
now also fixes #230 and #231
further addresses #227

x/airdrop/keeper/proposal_handler.go
```diff
@@ -49,6 +49,12 @@ func HandleRegisterZoneDropProposal(ctx sdk.Context, k Keeper, p *types.Register
 			return fmt.Errorf("claim record %d, %w", i, err)
 		}
 
+		for j, ca := range cr.ActionsCompleted {
```
joe-bowman, line 52: Should this just be in cr.ActionsCompleted != nil - it should be set here at all right?

joe-bowman (approved):
lgtm

x/airdrop/keeper/proposal_handler.go
```diff
@@ -49,9 +49,11 @@ func HandleRegisterZoneDropProposal(ctx sdk.Context, k Keeper, p *types.Register
 			return fmt.Errorf("claim record %d, %w", i, err)
 		}
 
-		for j, ca := range cr.ActionsCompleted {
-			if ctx.BlockTime().Before(ca.CompleteTime) {
-				return fmt.Errorf("invalid zonedrop proposal claim record [%d]: completed action [%d] is in future", i, j)
+		if cr.ActionsCompleted != nil {
```
joe-bowman, line 52: even simpler than that surely? we shouldn't accept Proposals with already completed actions? that map should always be nil - thus we should error if it is non-nil?

zanicar: yes indeed, got lost with what I was trying to do...

x/participationrewards/keeper/hooks.go
```diff
@@ -6,14 +6,27 @@ import (
...
-		k.Logger(ctx).Info("defer...")
+	if epochNumber < epochsDeferred {
+		k.Logger(ctx).Info("defer...", "epoch", epochNumber)
+
+		// create snapshot of current intents for the next epoch boundary
+		// requires intents to be set, no intents no snapshot...
+		// further snapshots will be taken during
+		// ValidatorSelectionRewardsCallback;
+		for _, zone := range k.icsKeeper.AllZones(ctx) {
+			for _, di := range k.icsKeeper.AllOrdinalizedIntents(ctx, zone, false) {
```
joe-bowman, line 25: I was just looking at this.

Do we require snapshotted intents to be ordinalized? 

Presently, intents are store as normalized - but snapshotted are ordinalized - but I cannot see why this is. If this is intentional it should be documented; otherwise we should switch to `AllIntents()`.

zanicar: I think we started using ordinalized intents (sorted by weight) when we experienced some determinism issues in earlier testnets... safe to revert to AllIntents()?

joe-bowman: yes please - saves a whole bunch of KV lookups each epoch, and makes everything consistent.

joe-bowman (approved):
lgtm

PR #243, opened 2022-09-05
add missing type assertions and legacy handler

joe-bowman:
- missing type assertions on custom Msgs
- missing legacy handler
- remove superfluous `RegisterCodec()` method that is superceded by `RegisterLegacyAminoCodec()`


zanicar (approved):
lgtm

PR #245, opened 2022-09-05
bug: fix nil balance denom determination; fixes #244

### LINKED ISSUE #244: x/interchainstaking: nil balance returns invalid denom in AccountBalanceCallback
When we receive an `accountBalanceCallback` with a nil balance, the balance string is not returned - the entire return is nil (because the value does not exist in the store). We must infer the denom from the request key.

In #222, we update the code at the request of the auditors, to make the callback simpler, but introduced a bug in doing so, as the length of the address is not considered when parsing the request key.

joe-bowman:
- fixes #244 by ensuring we add the address length to the index when returning the denom substring.
- refactor the parsing function into it's own method
- add test


zanicar (approved):
lgtm

PR #235, opened 2022-09-01
Release/v0.6.5

joe-bowman:
- fix bug in redemption request
- performance fix for intent ordinalization
- validation fixes from audit

PR #246, opened 2022-09-05
Release/v0.6.6 hotfix.2

joe-bowman:
- revert a number of empty byte slice guards introduced in #235 
- re-add participationrewards burn permission temporarily


zanicar (approved):
lgtm

PR #248, opened 2022-09-06
enum-fix updated

zanicar:
- zoneDrop Actions still zero indexed;
- Action type -1 for use in zd.Actions;
- zd.Actions idx + 1 for use as Action type;
- InBounds updated;
- errors updated;


joe-bowman (approved):
lgtm

PR #250, opened 2022-09-14
x/interchainstaking/types: add further tests + fix index out of range bugs for empty Allocations

odeke-em:
Addresses parts of Orijtech Inc audit findings, by adding more tests and coverage and found 2 bugs that could cause crashes when empty allocations are used. Instead of runtime errors, returns the empty object, and added tests to check for this condition.


odeke-em:
Kindly pinging @joe-bowman @mkaczanowski and an FYI for @elias-orijtech @kirbyquerby

joe-bowman (approved):
lgtm.

PR #252, opened 2022-09-15
x/interchainstaking/keeper: test zones roundtrip + fix bug in GetZoneForPerformanceAccount

odeke-em:
Addresses part of Orijtech Inc's cybersecurity audit to add more tests. Adds tests for roundtripping zones: storing and retrieving multiple zones. While here found and fixed a bug with (Keeper).GetZoneForPerformanceAccount in which if a zone was stored without a PerformanceAddress set, the code could panic because it didn't check for nil values before dereferencing the associated field .Address.


joe-bowman (approved):
lgtm

PR #262, opened 2022-12-15
Reflect final build tags in version

chillyvee:
v1.0.0 - db based build tags set after ldflags, so final db tags are not included in the version --long output.

Delay final ldflags until all tags are collected.


joe-bowman (approved):
awesome, ty ser!

PR #263, opened 2022-12-15
Allow COSMOS_BUILD_OPTIONS to be set externally

chillyvee:
Allow users to set build options without manually editing the Makefile


joe-bowman (approved):
lgtm

PR #264, opened 2022-12-16
update wasmvm lib to v1.1.1

joe-bowman:
- wasmvm library downloaded and included by the docker container bumped to the correct version.


ajansari95 (approved):
LGTM

PR #265, opened 2022-12-19
Ajansari95/escrow module account

ajansari95:


joe-bowman (changes requested):
lgtm, but can you rebase on main please, so we don't get the Dockerfile changes in the PR.

(I think this may have been based on v1.0.0).

ajansari95:
> lgtm, but can you rebase on main please, so we don't get the Dockerfile changes in the PR.
> 
> (I think this may have been based on v1.0.0).

the release v1.1.0 is one commit behind master will rebase release branch

joe-bowman:
Oh yes it probably is that. Thank you!

PR #272, opened 2022-12-30
Release/v1.1.0

joe-bowman:
Release v1.1.0

- add ics escrow module account to fix issue with locked up qAssets not being burnable post-redemption
- fixes situation where zero weight intents can exist if weight for given intent is zero
- remove burn permission from pr and mint module accounts
- remove pre v1.0.0 upgrade handlers
- add verbose errors for redelegation and unbonding tx failures
- direct poolincentives to airdrop module account
- replace redemption rate hard cap with soft cap
- add chain restart migration to remove cosmoshub-4, close channels and burn minted qAtoms


x/interchainstaking/keeper/v1_1_upgrade_handler_test.go
```diff
@@ -0,0 +1,99 @@
...
+	ibcchanneltypes "github.com/cosmos/ibc-go/v5/modules/core/04-channel/types"
+	tmclienttypes "github.com/cosmos/ibc-go/v5/modules/light-clients/07-tendermint/types"
+	"github.com/ingenuity-build/quicksilver/app"
+	"github.com/ingenuity-build/quicksilver/utils"
+	"github.com/ingenuity-build/quicksilver/x/interchainstaking/keeper"
+	icskeeper "github.com/ingenuity-build/quicksilver/x/interchainstaking/keeper"
+	icstypes "github.com/ingenuity-build/quicksilver/x/interchainstaking/types"
+)
+
+func (suite *KeeperTestSuite) Testv11UpgradeHandler() {
```
ajansari95, line 18:
```suggestion
func (suite *KeeperTestSuite) Test_v11UpgradeHandler() {
```
Just a NIT

x/participationrewards/types/expected_keepers.go
```diff
@@ -22,8 +22,6 @@ type BankKeeper interface {
 	SendCoinsFromModuleToAccount(ctx sdk.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error
 	SendCoinsFromModuleToModule(ctx sdk.Context, senderModule, recipientModule string, amt sdk.Coins) error
 	MintCoins(ctx sdk.Context, name string, amt sdk.Coins) error
-	BurnCoins(ctx sdk.Context, name string, amt sdk.Coins) error
```
ajansari95, line 25: I am mostly sure this is not related to v1.1.0 but any reason?

ajansari95 (approved):
Just a NIT, otherwise LGTM

x/participationrewards/types/expected_keepers.go
joe-bowman: This was removed from expected keepers, because it was only used during dev (before all elements of PR was completed), and I just removed the permissions on PR module account to burn coins; it is no longer used, or usable, in the PR module.

ajansari95 (approved):

x/interchainstaking/keeper/intent.go
```diff
@@ -136,14 +136,17 @@ func (k *Keeper) AggregateIntents(ctx sdk.Context, zone *types.Zone) error {
 	}
 
 	// normalise aggregated intents again.
-	for idx, intent := range aggregate.Sort() {
-		intent.Weight = intent.Weight.Quo(ordinalizedIntentSum)
-		aggregate[idx] = intent
+	newAggregate := make(types.ValidatorIntents, 0)
```
zanicar, line 139: Creating this with `len(aggregate)` and using an idx to set the values is much more performant...
Go slice allocation is rather slow, so if this is being used a lot setting it to the correct size initially will be very beneficial.

zanicar (commented):
Added a performance improvement note... as this may be a good time for such a simple enhancement.

x/interchainstaking/keeper/intent.go
joe-bowman: We don't know the correct size ahead of time, as we are filtering out zero values.

Performance is less critical as this get called once per epoch.

PR #274, opened 2023-01-02
sort account map prior to burning; non-determinism!

joe-bowman:
- sort address map before burning, to avoid non-determinism


ajansari95 (approved):

zanicar (approved):
lgtm

PR #282, opened 2023-01-18
Release/v1.2.x

joe-bowman:
## 1. Summary
Fixes:
- redemption rate calculation race conditions caused by paginated query response.

## 2.Type of change

## 3. Implementation details

- Instead of deleting delegation records on `delegations` callback, we take no action until the provable `delegation` callback, which avoids the deletion and subsequent re-adding of delegations at the epoch boundary when the delegations query is paginated.

## 4. How to test/use

Submit an incomplete / paginated delegatorDelegations query IBC response to the `delegations` callback; delegation records should not be deleted.

## 5. Checklist

- [x] Have tests been added to cover the change

## 6. Limitations (optional)

N/A

## 7. Future Work (optional)

N/A


ajansari95 (approved):
LGTM

PR #295, opened 2023-02-15
chore: fix `go version` check in `Makefile`

aljo242:
## 1. Summary

The existing `check_version` target in our `Makefile` ONLY supported `go v1.19.x`. This PR changes the logic to check if the go version is greater than or equal to `v1.19`, allowing for `v1.20.x` and above to be used.

NOTE: 

Since we already are requiring `go 1.19` in our `go.mod` file, I'm not actually sure how necessary this check is.  All compilation  / testing / etc. will need to parse and check the `go.mod` file anyways.

## 2.Type of change

## 4. How to test/use

Verify that build works with go `v1.19.x` and `v1.20.x` while `v1.18.x` and below are NOT supported.
## 5. Checklist

## 6. Limitations (optional)

Verify that there are no consensus breaking changes made from `1.19` to `1.20`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ endif
 
 check_version:
 ifneq ($(shell expr $(GO_MINOR_VERSION) \>= 19), 1)
-	@echo "ERROR: Go version 1.19 is required for building Quicksilver. There are consensus breaking changes between binaries compiled with Go 1.18 and Go 1.19."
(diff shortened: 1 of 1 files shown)


joe-bowman:
We should force operators to use a single Go minor version. For v1.4.0 this should be go1.19 and for v1.5.0 this will be go1.20. I think we can close this PR in this case?

aljo242:
Agreed - closing now

PR #302, opened 2023-02-17
fix: flaky test with `go 1.20`

aljo242:
## 1. Summary

The following [test](https://github.com/ingenuity-build/quicksilver/blob/0bf2733f875074a5c4941ebf40fc04d6d6afc8e5/x/interchainstaking/keeper/ibc_packet_handlers_test.go#L1253) was failing because of an inconsistency in `sdk.MustDecFromStr()` that only occurs in `go v1.20`.  Converting a `float32` instead of a `float64` into an `sdk.Dec` solves this issue, but I am unsure why this is the case given that it is only occurring in `v1.20`.  Making this modification does not make the tests any less strong, so I still recommend we make it.

## 2.Type of change


joe-bowman (approved):
lgtm

ajansari95 (approved):
LGTM

PR #310, opened 2023-02-20
chore: add `tools.go`

aljo242:
## 1. Summary

Adds a `tools` pkg which allows us to require development tools like `golangci-lint` and `gofumpt` in our `go.mod` file.  This is potentially useful for new devs working with our code who may not have the tools installed.  Otherwise, they would receive a `make` target fail if they tried to run something like `make format` without the tool installed.

This addition also lets us ensure that all developers are using the same minimum version for said tools.

## 2.Type of change

## 3. Implementation details

- add `tools/tools.go` 
- Use build tags to ensure these are not built into our binaries

## 4. How to test/use

- run `go clean -modcache` and remove tools like `gofumpt`
- run `go mod tidy` and then try to use tools again


ajansari95 (approved):
LGTM

joe-bowman (approved):
lgtm

ajansari95 (approved):
LGTM

PR #313, opened 2023-02-21
added responseTypecheck

ajansari95:
## 1. Summary
Fixes # (issue)
- msgResponse which is 0 byte is also not nil
- have a check on msgResponseType if its not empty string then 46 otherwise 45

## 2.Type of change

## 5. Checklist

### DIFF
--- a/x/interchainstaking/keeper/ibc_packet_handlers_test.go
+++ b/x/interchainstaking/keeper/ibc_packet_handlers_test.go
@@ -3,6 +3,7 @@ package keeper_test
 import (
 	"crypto/sha256"
 	"fmt"
+	distrtypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
 	"math/rand"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


x/interchainstaking/keeper/ibc_packet_handlers_test.go
```diff
@@ -1618,3 +1619,116 @@ func (s *KeeperTestSuite) Test_v046Callback() {
 	// assert that fee collector module balance is now 100denom more than before HandleMsgTransfer()
 	s.Require().Equal(feeMaccBalance2.AmountOf("denom").Sub(feeMaccBalance.AmountOf("denom")), sdk.NewInt(100))
 }
+
+func (s *KeeperTestSuite) Test_v045Callback_NilResp() {
+	s.SetupTest()
```
aljo242, line 1624: logic is duplicated between both tests.  Let's make it a table-driven test to make things a bit simpler and clearer.

aljo242 (changes requested):

x/interchainstaking/keeper/ibc_packet_handlers_test.go
ajansari95: check #316

PR #316, opened 2023-02-22
fix issue with negative available per validator

joe-bowman:
## 1. Summary
- add guards against negative completion times (should not be possible now, but belt and braces)
- panic (hooks and tx, so safe) on negative allocations
- fixes a test that no longer generates an error, given no satisfiable unbondings
- adds a check on msgResponseType if its not empty string then 46 otherwise 45

## 2.Type of change

## 3. Implementation details

Add guards against invalid completion times.
Panic on negative unbonding allocation.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -26,6 +26,7 @@ const (
 	v010400UpgradeName    = "v1.4.0"
 	v010400rc6UpgradeName = "v1.4.0-rc6"
 	v010400rc7UpgradeName = "v1.4.0-rc7"
+	v010400rc8UpgradeName = "v1.4.0-rc8"
 )
 
 func isTest(ctx sdk.Context) bool {
@@ -50,6 +51,7 @@ func setUpgradeHandlers(app *Quicksilver) {
 	app.UpgradeKeeper.SetUpgradeHandler(v010400UpgradeName, v010400UpgradeHandler(app))
 	app.UpgradeKeeper.SetUpgradeHandler(v010400rc6UpgradeName, v010400rc6UpgradeHandler(app))
 	app.UpgradeKeeper.SetUpgradeHandler(v010400rc7UpgradeName, noOpHandler(app))
+	app.UpgradeKeeper.SetUpgradeHandler(v010400rc8UpgradeName, v010400rc8UpgradeHandler(app))
 
 	// When a planned update height is reached, the old binary will panic
 	// writing on disk the height and name of the update that triggered it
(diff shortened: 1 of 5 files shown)


ajansari95:
Closes comments from #313 for table based check and closes the PR

x/interchainstaking/keeper/redemptions.go
```diff
@@ -304,6 +308,7 @@ func DetermineAllocationsForUndelegation(currentAllocations map[string]math.Int,
 	for idx := range deltas {
 		deltas[idx].Weight = deltas[idx].Weight.Sub(sdk.NewDecFromInt(maxValue)).Abs()
 		sum = sum.Add(deltas[idx].Weight.TruncateInt().Abs())
+		fmt.Printf("Dropped Val: %s, Weight: %s\n", deltas[idx].ValoperAddress, deltas[idx].Weight.TruncateInt())
```
joe-bowman, line 311: Can we remove these extra log lines please

x/interchainstaking/keeper/redemptions.go
```diff
@@ -316,6 +321,9 @@ func DetermineAllocationsForUndelegation(currentAllocations map[string]math.Int,
 			if !ok {
 				availablePerValidator[deltas[idx].ValoperAddress] = sdk.ZeroInt()
 			}
+			fmt.Println("Raw allocation", allocation)
```
joe-bowman, line 324: Add her

joe-bowman (commented):
Minor bits but otherwise good

x/interchainstaking/keeper/redemptions.go
ajansari95: do we want to remove all fmt.Println here?

ajansari95: i can see more

app/upgrades.go
```diff
@@ -198,3 +200,29 @@ func v010400rc6UpgradeHandler(app *Quicksilver) upgradetypes.UpgradeHandler {
...
+				}
+				return false
+			})
+			return false
+		})
+
+		app.InterchainstakingKeeper.IterateRedelegationRecords(ctx, func(_ int64, key []byte, record icstypes.RedelegationRecord) (stop bool) {
+			if record.CompletionTime.Unix() <= 0 {
+				app.InterchainstakingKeeper.Logger(ctx).Info("Removing delegation record", "chainid", record.ChainId, "source", record.Source, "destination", record.Destination, "epoch", record.EpochNumber)
+				app.InterchainstakingKeeper.DeleteRedelegationRecord(ctx, record.ChainId, record.Source, record.Destination, record.EpochNumber)
```
aljo242, line 221: Currently not testing this branch of the `if`.  Otherwise, lgtm with @joe-bowman's comments addressed.

ajansari95: actually it does the test L153:uprades_test.go - I resued the earlier state to test ; )

aljo242 (approved):

ajansari95 (approved):
LGTM

PR #321, opened 2023-02-24
refactor: memo parsing logic in `ibc_packet_handlers.go`

aljo242:
## 1. Summary

refactor duplicated logic in `x/interchainstaking/keeper/ibc_packet_handlers.go` into a `types` package.

## 2.Type of change

## 3. Implementation details

- Refactor duplicated code into `types` package
- add tests
- add fuzz tests

## 4. How to test/use

To run the fuzz test:

```
go test -fuzz=FuzzParseMsgMemo github.com/ingenuity-build/quicksilver/x/interchainstaking/types -fuzztime=5m 
```

## 5. Checklist


joe-bowman (commented):
Looks good overall, but as discussed there seems to be no difference between the fuzz test and the unit test here in terms of coverage. As fuzz tests are not currently run on every push, and the unit test is not resource intensive, should be just keep that?

ajansari95:
> Looks good overall, but as discussed there seems to be no difference between the fuzz test and the unit test here in terms of coverage. As fuzz tests are not currently run on every push, and the unit test is not resource intensive, should be just keep that?

i think we can add a fuzz test target in make file and not include in coverage?

joe-bowman (approved):
lgtm

ajansari95 (approved):
LGTM

PR #331, opened 2023-03-03
refactor: use constant for redemptions GC interval

aljo242:
## 1. Summary

Using a constant is more clear.  Do we want this to be a static constant or something that has more control like a param?


joe-bowman (approved):
lgtm

ajansari95 (approved):
LGTM

PR #334, opened 2023-03-03
chore: workflow improvements

aljo242:
## 1. Summary

Add two new workflows that are both run once a week:
- long sim test (runs full `make-test-sim-non-determinism`)
- markdown link checker (lets us know if any links break over time)

Removes the `push` tag for the `buildtest` workflow.  This was  causing workflow duplicates to be run.

Adds build and test rules to run on `darwin/arm64` (apple silicon) and `darwin/amd64`


.github/workflows/simulation.yml
```diff
@@ -1,17 +1,14 @@
 name: simulate quicksilver
 
 on:
-  pull_request:
-    branches:
-      - main
+  schedule:
+    - cron: '0 0 * * 0' # run once a week
```
ajansari95, line 5: can we do it based on some trigger? if we want to force run on some pe?

aljo242: So I moved the `run-sim` label trigger to `buildtest.yaml`.

This new scheduled job is an additional run of a really long simulation (more blocks and larger blocksize).

aljo242: Trigger still exists, just got moved to another file

joe-bowman (dismissed):
lgtm

ajansari95 (approved):
LGTM

joe-bowman (approved):
lgtm

PR #349, opened 2023-03-14
backport: gov reopen/close to `v1.2.5`

aljo242:
## 1. Summary

Fixes QCK-174

- Backport `GovReopenChannel` and `GovCloseChannel`

- Add `InterchainStakingKeeper.SetEpochsKeeper()` to prevent a panic

Note: we should add test for the `msg_server` functionality in the future.


x/interchainstaking/keeper/callbacks_test.go
```diff
@@ -588,14 +589,17 @@ func (s *KeeperTestSuite) TestHandleRewardsCallbackEmptyResponse() {
 
 		respbz, err := app.AppCodec().Marshal(&response)
 		s.Require().NoError(err)
+
+		epoch := app.EpochsKeeper.GetEpochInfo(ctx, "epoch")
+		fmt.Println(epoch)
```
joe-bowman, line 594: Debug

x/interchainstaking/keeper/keeper.go
```diff
@@ -634,3 +639,7 @@ func DetermineAllocationsForRebalancing(currentAllocations map[string]math.Int,
 
 	return out
 }
+
```
joe-bowman, line 642: If we are adding setEpochsKeeper, do we need to remove epochkeeper from NewKeeper (I added in v1.2.5)

x/interchainstaking/keeper/msg_server.go
```diff
@@ -317,3 +320,86 @@ func (k Keeper) EmitValsetRequery(ctx sdk.Context, connectionID string, chainID
 	)
 	return nil
 }
+
+// GovReopenChannel reopens an ICA channel.
+func (k msgServer) GovReopenChannel(goCtx context.Context, msg *types.MsgGovReopenChannel) (*types.MsgGovReopenChannelResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// checking msg authority is the gov module address
```
joe-bowman, line 328: This is no longer a proposal, so can remove these commented lines

joe-bowman (changes requested):
A couple of comments - can you also add the cli tx command for reopen

PR #369, opened 2023-03-29
chore: use `develop` in workflows and coverage

aljo242:
When we switched `main` and `develop` we left some changes in workflows that needed to be switched over:
- base main CodeCov report off of `develop` instead of `main`
- do workflows against pushes to `main` AND `develop`


ajansari95 (commented):
it should run on both PR's targetted to develop and main. And also if we want to merge to a feature/xxx branch if possible?

aljo242:
> it should run on both PR's targetted to develop and main. And also if we want to merge to a feature/xxx branch if possible?

So it is run against ALL PRs, the rules were just for workflow runs when pushed to branches were made.

ajansari95 (approved):
LGTM

PR #370, opened 2023-03-30
fix: error code for running binary

aljo242:
Fixes error where 1 (error) was returned on all runs of the binary


ajansari95 (approved):
lgtm

joe-bowman (approved):
ship it

PR #375, opened 2023-04-03
chore: add staticmajor analysis tool

aljo242:
## 1. Summary

Add `staticmajor` tool as a gh workflow.  Configured to run all checks and ignore proto generated files.

### DIFF
--- a/.github/workflows/staticmajor.yml
+++ b/.github/workflows/staticmajor.yml
@@ -1,0 +1,17 @@
+name: Staticmajor Analysis
+on: [push,workflow_dispatch,pull_request]
+jobs:
+  run_staticmajor:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Staticmajor action
+        id: staticmajor
+        uses: orijtech/staticmajor-action@main
+        with:
+          resleak: true
+          httperroryzer: true
+          structslop: true
+          fuzzy: true
+          generated: false


elias-orijtech:
Ok, here's an example run of the staticmajor checker on my quicksilver fork:
https://github.com/elias-orijtech/quicksilver/actions/runs/4631076506/jobs/8193493524

Two things:
(1) Quicksilver doesn't build with CGO_ENABLED=0 leading to a bunch of noise from httperroryzer and structslop.
(2) The generated flag is only implemented for the resleak analyzer, so .pb.go files are analyzed by httperroryzer and structslop.

For (1), I assume the noise can be suppressed in `httperroryzer` and `structslop` but the best approach is to support CGO_ENABLED=0 builds so that analyzers don't miss any code.
For (2) you can either disable `httperroryzer` and `structslop` or we can implement `generated=false` in the analyzers if there's enough interest to keep them enabled.

I can work on either as needed; let me know.

joe-bowman (approved):
lgtm

ajansari95 (approved):

odeke-em:
@elias-orijtech let's perhaps work on extending skipping of *.pb.go filed for httperroyzer and structslop, that's the more scalable approach that's gonna work for more repos and least work for Quicksilver. Thank you

elias-orijtech:
Skipping the generated files is easy, but I don't see a practical way around the CGO_ENABLED=0 issue. Answered in the Slack thread.

PR #382, opened 2023-04-05
test: `interchainstaking` zones cli query

aljo242:
## 1. Summary

Set up `TestGetCmdZonesInfos()` so that there is an initial state with zones to be queried for instead of an empty state.

## 2.Type of change

- [X] Test

## 7. Future Work (optional)

Frontport to `develop`


joe-bowman (approved):
lgtm

ajansari95 (approved):
LGTM

PR #387, opened 2023-04-07
enhancement: validators struct separated

ajansari95:
Fixes QCK-285

- Perf improvement for zones struct
- separate out validators struct
- lookup for validators based on chain-id


ajansari95:
testing suite setup needs to be fixed tests failing

joe-bowman:
Would this be more performant to store each validator under it's own key. We do a bunch of operations on single validators, and fetching the entire valset for each is counterintuitive there. We can still fetch all validators for a chain by prefixing the key with the chain-id (make key chain-id + validatorbytes (not bech32 prefixed address, because 51 bytes vs 20 bytes))

aljo242:
Make sure to run https://github.com/ingenuity-build/quicksilver/pull/389 targets once merged

joe-bowman:
Closing in favour of #398

PR #394, opened 2023-04-13
perf: use an address <-> zone mapping to avoid heavy lookups

joe-bowman:
## 1. Summary
Removes existing logic that caused multiple iterations over zones in order to map an ICA address to a zone.
Introduces a mapping between address and chain id to perform more performant lookups.

## 2.Type of change

- [x] Performance improvement

## 3. Implementation details

Introduces a mapping between address and chain id to perform more performant lookups.

## 4. How to test/use

N/A covered by existing tests.

## 5. Checklist

## 6. Limitations (optional)

This does not make zone fetches more performant, but reduces the number of them.

## 7. Future Work (optional)

We still need to make improvements to the zone struct to reduce spiralling gas costs.


aljo242 (approved):
LGTM - let's add backlog for testing the new functions

aljo242 (approved):

PR #400, opened 2023-04-18
fix(workflows): remove static major

aljo242:
Remove static major github workflow as it is not compatible when `CGO_ENABLED=0` does not work. 

I believe we cannot have `CGO_ENABLED=0` fully work because of wasm requirements.


joe-bowman:
sad times :'(

joe-bowman (approved):
ack

ajansari95 (approved):

PR #408, opened 2023-05-02
feat: `x/airdrop` IncentivePoolSpend 

aljo242:
## 1. Summary
Fixes QCK-327

Add `IncentivePoolSpend` which  which can only be executed via governance.  Sends tokens from the `airdrop` module account to a given address.

## 2.Type of change

## 3. Implementation details

- Add the `authority` field to the `airdrop` keeper.  This field is given in the `app` package as the `gov` module account name.  This means that the app can check if a message sent and signed by the proper authority and will only execute it if the authority is correct.

## 4. How to test/use

TODO - add `msg_server`-level test

## 5. Checklist


x/airdrop/keeper/msg_server.go
```diff
@@ -32,3 +38,42 @@ func (k msgServer) Claim(goCtx context.Context, msg *types.MsgClaim) (*types.Msg
...
+			if a.Amount.IsInt64() {
+				telemetry.SetGaugeWithLabels(
+					[]string{"tx", "msg", "send"},
+					float32(a.Amount.Int64()),
+					[]metrics.Label{telemetry.NewLabel("denom", a.Denom)},
+				)
+			}
+		}
+	}()
+
```
joe-bowman, line 77: Do we need any events emitted here?

x/airdrop/types/msgs.go
```diff
@@ -84,3 +89,58 @@ func (msg MsgClaim) GetSigners() []sdk.AccAddress {
...
+		return sdkerrors.ErrInvalidAddress.Wrapf("invalid from address: %s", err)
+	}
+
+	to, err := sdk.AccAddressFromBech32(msg.ToAddress)
+	if err != nil {
+		return sdkerrors.ErrInvalidAddress.Wrapf("invalid to address: %s", err)
+	}
+
+	if from.Equals(to) {
+		return sdkerrors.ErrInvalidAddress.Wrapf("to and from addresses equaul: %s", err)
```
joe-bowman, line 123:
```suggestion
		return sdkerrors.ErrInvalidAddress.Wrapf("to and from addresses equal: %s", err)
```

x/airdrop/keeper/msg_server.go
aljo242: SendCoinsFromModuleToAccount will call [SendCoins](https://github.com/cosmos/cosmos-sdk/blob/5cd0b2316a7103468af38eab5d886f9f069c9cd7/x/bank/keeper/send.go#L135) which will emit the basic send events already.  I don't think we need to include any other information

joe-bowman: Ah yes, perfect, ty.

joe-bowman (approved):
lgtm

ajansari95 (approved):
lgtm

PR #409, opened 2023-05-02
fix: codecov uploads

aljo242:
## 1. Summary
Fixes QCK-328

Upload reports on pushes to `main` and `develop` so we can compare against base coverages.


joe-bowman (approved):
lgtm

ajansari95 (approved):

PR #425, opened 2023-05-11
draft ADR-001 for RS rewards

joe-bowman:
- add ADR-001 for multidenomination rewards


architecture/adr-001-multi-denomination-rewards.md
```diff
@@ -0,0 +1,88 @@
...
+
+The current implementation of rewards distribution is such that
+rewards collected in the host zone's staking denom are delegated, minus a
+parameterized fee that is sent, along with the remaining rewards (any non-staking-
+denom fees that have been collected) to the FeeCollectorAccount on Quicksilver to
+be distributed to QCK token holders.
+
+With rewards being accrued in multiple denominations, this proposal asserts that
+these rewards should no longer be directed to QCK token holder, but to the holders
+of the qAssets whose underlying capital generated the rewards.
```
aljo242, line 27:
```suggestion
With rewards being accrued in multiple denominations, this proposal asserts that
these rewards should no longer be directed to _all_ QCK token holders, but to the holders
of the qAssets whose underlying capital generated the rewards.
```

@joe-bowman this is what you mean right?

How heavyweight will it be to track all of this?

joe-bowman: No different to participation rewards distribution (+qs2)

architecture/adr-001-multi-denomination-rewards.md
```diff
@@ -0,0 +1,88 @@
...
+With the advent of replicated security, delegators now accrue rewards in multiple
+denominations. These collected rewards should be distributed proportionally to the
+appropriate qAsset holders for the host zone.
+
+## Context
+
+The current implementation of rewards distribution is such that
+rewards collected in the host zone's staking denom are delegated, minus a
+parameterized fee that is sent, along with the remaining rewards (any non-staking-
+denom fees that have been collected) to the FeeCollectorAccount on Quicksilver to
```
ajansari95, line 22: do we move the non-staking denoms to a different account and then distribute?

joe-bowman: Yes; the partial implementation [here](https://github.com/ingenuity-build/quicksilver/pull/427) uses the local account that corresponds to the withdrawal address. This saves us having to store another field, and is a value we know will not change over time.

PR #428, opened 2023-05-11
fix: QCK-425 update delegations on redelegation

joe-bowman:
## 1. Summary
Fixes # QCK-425

Avoid panic at epoch by ensuring redelegations cause update of underlying delegation records for both source and dst validators.

## 2.Type of change

## 3. Implementation details

When MsgBeginRedelegate acknowledge is received, we now update both the source and destination validator delegation records based upon the amount that was redelegated. Additionally, we trigger a KV store lookup against both delegations to ensure they are accurate against the host chain. 

The reason we do both is to avoid potential race conditions, where a redelegation might be acknowledged, and we use the delegation message immediately after before the ICQ request has an opportunity to be updated.

## 4. How to test/use

Trigger a redelegation at an epoch boundary (by changing intent, for example), and check the two delegation records involved reflect the redelegation change.

### DIFF
--- a/app/upgrades/types.go
+++ b/app/upgrades/types.go
@@ -22,6 +22,7 @@ const (
 	V010402rc3UpgradeName = "v1.4.2-rc3"
 	V010402rc4UpgradeName = "v1.4.2-rc4"
 	V010402rc5UpgradeName = "v1.4.2-rc5"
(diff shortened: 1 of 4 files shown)


ajansari95 (approved):
Lgtm

ajansari95 (dismissed):

PR #433, opened 2023-05-16
bug fix in checking receipts and using safesub

ajansari95:
## 1. Summary
Fixes # (issue)

- Fix for case when `delegationAccountBalance.Sub(amount)` is negative i.e we have less amount of pending delegations then the amount from aggregating receipts 

## 2.Type of change

## 3. Implementation details

- using `safeSub` instead of `Sub` to avoid panic  

## 5. Checklist

### DIFF
--- a/x/interchainstaking/keeper/delegation.go
+++ b/x/interchainstaking/keeper/delegation.go
@@ -415,15 +415,14 @@ func (k *Keeper) FlushOutstandingDelegations(ctx sdk.Context, zone *types.Zone, 
 	exclusionTime := ctx.BlockTime().AddDate(0, 0, -1)
 
 	k.IterateZoneReceipts(ctx, zone, func(_ int64, receiptInfo types.Receipt) (stop bool) {
-		if receiptInfo.FirstSeen.After(exclusionTime) && receiptInfo.Completed == nil {
+		if receiptInfo.FirstSeen.Before(exclusionTime) && receiptInfo.Completed == nil {
 			amount = amount.Add(receiptInfo.Amount...)
 		}
 		return false
 	})
 
-	coins := sdk.NewCoins(delAddrBalance).Sub(amount...)
-
-	if coins.IsAnyNegative() || coins.IsZero() {
+	coins, hasNeg := sdk.NewCoins(delAddrBalance).SafeSub(amount...)
+	if hasNeg || coins.IsZero() {
 		k.Logger(ctx).Debug("delegate account balance negative, setting outdated reciepts")
 		k.SetReceiptsCompleted(ctx, zone, exclusionTime, ctx.BlockTime())
 		return nil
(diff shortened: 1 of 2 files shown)


joe-bowman (approved):
lgtm

.github/workflows/golangci.yml
```diff
@@ -17,9 +17,10 @@ jobs:
     steps:
       - uses: actions/setup-go@v4
         with:
-          go-version: 1.19
+          go-version: 1.20.2
```
joe-bowman, line 20: should this be updated? main should still be on 1.19

ajansari95: nope this is just for the lint version

x/interchainstaking/keeper/delegation.go
```diff
@@ -413,17 +413,31 @@ func (k *Keeper) MakePerformanceDelegation(ctx sdk.Context, zone *types.Zone, va
...
+			return false
+		})
+
+		coins := sdk.NewCoins(delAddrBalance).Sub(amount...)
+		if coins.IsAnyNegative() || coins.IsZero() {
+			k.Logger(ctx).Debug("delegate account balance negative, setting outdated reciepts")
+			k.SetReceiptsCompleted(ctx, zone, exclusionTime, ctx.BlockTime())
+			return nil
+		}
+	}
```
joe-bowman, line 430: should this not be `else { ... new logic ... }` ?

joe-bowman, line 430: should this not be `else { ... new logic ... }` ?

joe-bowman: otherwise `coins` below will get clobbered, right?

ajansari95: no my bad

joe-bowman: coins is positive? that will fall through to the new logic.

joe-bowman: in which case l424 shouldn't have `:`, because else it'll create a new val scoped within that if statement.

x/interchainstaking/keeper/delegation.go
```diff
@@ -413,20 +412,36 @@ func (k *Keeper) MakePerformanceDelegation(ctx sdk.Context, zone *types.Zone, va
...
+	} else {
+		k.IterateZoneReceipts(ctx, zone, func(_ int64, receiptInfo types.Receipt) (stop bool) {
+			if receiptInfo.FirstSeen.Before(exclusionTime) && receiptInfo.Completed == nil {
+				amount = amount.Add(receiptInfo.Amount...)
+			}
+			return false
+		})
+
+		coins2, hasNeg := sdk.NewCoins(delAddrBalance).SafeSub(amount...)
+		coins = coins2
```
joe-bowman, line 439:
```suggestion
		var hasNeg bool
		coins, hasNeg = sdk.NewCoins(delAddrBalance).SafeSub(amount...)
```

joe-bowman: would be my preference, but simply stylistic (and saves an additional allocation)

joe-bowman (dismissed):
lgtm

PR #451, opened 2023-06-08
adding ibc hooks

muku314115:
## 1. Summary
Areas that need extra attention: 
- go.mod file has not been added separately to ibc-hooks as is done in osmosis - open to suggestions
- wasm_hooks.go - changes made to accomodate the updates present in ibc-go/v7
- app/modules.go - ibc-hooks do not contain any end/begin blockers but the order set for them might be worth checking out

## 2.Type of change

## 5. Checklist


ajansari95:
@muku314115  can we change the base branch to `unstable-v1.5` i think its unblocked now

PR #420, opened 2023-05-10
Sdk47 rightway

faddat:
## 1. Summary
Fixes # (issue)

## 2.Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## 5. Checklist

- [ ] Does the Readme need to be updated?

PR #464, opened 2023-06-18
feat: subzones part 1

aljo242:
## 1. Summary

Subzones functionality part 1.  Splitting this PR into multiple parts because it is getting too large and I think its better to get some feedback at this intermediate stage

## 2.Type of change

## 3. Implementation details

- add `SubzoneInfo` to `Zone` struct
- add `SubzoneInfo` helper functions
- add subzone creation logic to the zone proposal handler
- remove all direct usages of `zone.ChainId` and replace with the appropriate helper functions

## 5. Checklist

## 7. Future Work (optional)

Part 2:
- more testing (we need to create tests for `HandleRegisterZoneProposal` in general - not  just the subzone creation case)
- handle minting and control for subzone authority


proto/quicksilver/interchainstaking/v1/interchainstaking.proto
```diff
@@ -49,6 +49,13 @@ message Zone {
...
   bool return_to_sender = 27;
   bool is_118 = 28;
+  SubzoneInfo subzoneInfo = 29;
+}
+
+message SubzoneInfo {
+  string authority = 1;
+  string base_chainID = 2;
+  string chainID = 3;
 }
```
aljo242, line 59: needed to add `chainID` here so that we pass this value into the `RegisterZoneProposal` as a unique identifier that is not generated from the `connectionID`

x/interchainstaking/keeper/callbacks_test.go
```diff
@@ -347,11 +347,11 @@ func (suite *KeeperTestSuite) TestHandleValidatorCallbackNilValue() {
...
-	suite.GetQuicksilverApp(suite.chainA).InterchainstakingKeeper.SetValidator(suite.chainA.GetContext(), zone.ChainId, icstypes.Validator{ValoperAddress: "cosmosvaloper1a3yjj7d3qnx4spgvjcwjq9cw9snrrrhu5h6jll", CommissionRate: sdk.MustNewDecFromStr("0.2"), VotingPower: sdk.NewInt(2000), DelegatorShares: sdk.NewDec(2000)})
-	suite.GetQuicksilverApp(suite.chainA).InterchainstakingKeeper.SetValidator(suite.chainA.GetContext(), zone.ChainId, icstypes.Validator{ValoperAddress: "cosmosvaloper1z8zjv3lntpwxua0rtpvgrcwl0nm0tltgpgs6l7", CommissionRate: sdk.MustNewDecFromStr("0.2"), VotingPower: sdk.NewInt(2000), DelegatorShares: sdk.NewDec(2000)})
+	suite.GetQuicksilverApp(suite.chainA).InterchainstakingKeeper.SetValidator(suite.chainA.GetContext(), zone.ChainID(), icstypes.Validator{ValoperAddress: "cosmosvaloper1sjllsnramtg3ewxqwwrwjxfgc4n4ef9u2lcnj0", CommissionRate: sdk.MustNewDecFromStr("0.2"), VotingPower: sdk.NewInt(2000), DelegatorShares: sdk.NewDec(2000)})
```
aljo242, line 350: Using `ChainID()` when we are accessing common data between base zones and sub zones (validators, etc).  We use `ID()` when accessing / storing anything that needs be unique between zones (receipts, etc.)

aljo242: There are a lot of replacement strings just for the *ID() functions, so please help me verify that I replaced properly

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -619,7 +621,7 @@ func (k *Keeper) HandleBeginRedelegate(ctx sdk.Context, msg sdk.Msg, completion
 	k.ICQKeeper.MakeRequest(
 		ctx,
 		zone.ConnectionId,
-		zone.ChainId,
+		zone.ChainID(),
```
joe-bowman, line 624: I think this wants to be ID() - reason being, the query callback will be executed against the zone listed here, so and the delegation belongs to the subzone, not the rootzone.

joe-bowman: Same goes for receipts, balances, withdrawal / unbending records, etc. (not valsets and validators, or performance account delegations), as these will always pertain to the subzone.

icq relayer will need to be updated to split the chain ID on the separator (probably: `/` ?) and use the correct base chain id for the query itself.

joe-bowman:
I think all queries with the exception of performance account,l valet and validator queries should be targeted at the ChainID(), not ID().

x/interchainstaking/keeper/ibc_packet_handlers.go
aljo242: Ok - so we will want to ensure some deterministic way to get the base chainID from a subzone chainID right?

Should we just generate subzone ID as `baseID + "/something that won't conflict"?

aljo242:
> MakeRequest

@joe-bowman what about balance queries?

aljo242:
just codecov upload fail

x/interchainstaking/keeper/ibc_packet_handlers.go
joe-bowman: As discussed yesterday `baseChainID + "|" + <user-generated unique value per baseChain>`

ajansari95 (commented):
conflicts

x/interchainstaking/client/cli/tx.go
```diff
@@ -174,8 +175,21 @@ Where proposal.json contains:
...
+				proposal.LocalDenom,
+				proposal.AccountPrefix,
+				proposal.ReturnToSender,
+				proposal.UnbondingEnabled,
+				proposal.DepositsEnabled,
+				proposal.LiquidityModule,
+				proposal.Decimals,
+				proposal.MessagesPerTx,
+				proposal.Is_118,
+			)
```
ajansari95, line 192: handle subzone here?

x/interchainstaking/keeper/receipt_test.go
```diff
@@ -78,7 +78,7 @@ func (suite *KeeperTestSuite) TestHandleReceiptTransactionBadRecipient() {
 
 	err = icsKeeper.HandleReceiptTransaction(ctx, transaction, hash, zone)
 	// suite.Require().ErrorContains(err, "no sender found. Ignoring")
-	nilReceipt, found := icsKeeper.GetReceipt(ctx, types.GetReceiptKey(zone.ChainId, hash))
+	nilReceipt, found := icsKeeper.GetReceipt(ctx, types.GetReceiptKey(zone.ChainID(), hash))
```
muku314115, line 81: Should this be zone.ID()

ajansari95 (dismissed):
lgtm

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -1115,7 +1116,7 @@ func (k *Keeper) TriggerRedemptionRate(ctx sdk.Context, zone *types.Zone) error
...
```
joe-bowman, line 1119: AllBalances on SubZone -> `ID()`

x/interchainstaking/keeper/keeper.go
```diff
@@ -434,7 +434,7 @@ func (k *Keeper) EmitPerformanceBalanceQuery(ctx sdk.Context, zone *types.Zone)
 	k.ICQKeeper.MakeRequest(
 		ctx,
 		zone.ConnectionId,
-		zone.ChainId,
+		zone.ID(),
```
joe-bowman, line 437: Performance account only uses by Base Chain - `ChainID()` here

x/interchainstaking/keeper/keeper.go
```diff
@@ -497,7 +497,7 @@ func (k *Keeper) EmitDepositIntervalQuery(ctx sdk.Context, zone *types.Zone) {
 	k.ICQKeeper.MakeRequest(
 		ctx,
 		zone.ConnectionId,
-		zone.ChainId,
+		zone.ChainID(),
```
joe-bowman, line 500: `ID()`

x/interchainstaking/keeper/keeper.go
```diff
@@ -578,7 +578,7 @@ func (k *Keeper) GetAggregateIntentOrDefault(ctx sdk.Context, z *types.Zone) (ty
 	var filteredIntents types.ValidatorIntents
 
 	if len(z.AggregateIntent) == 0 {
-		intents = k.DefaultAggregateIntents(ctx, z.ChainId)
+		intents = k.DefaultAggregateIntents(ctx, z.ChainID())
```
joe-bowman, line 581: `ID()`

x/interchainstaking/keeper/proposal_handler.go
```diff
@@ -16,16 +16,37 @@ import (
...
+
 	// get chain id from connection
 	chainID, err := k.GetChainID(ctx, p.ConnectionId)
 	if err != nil {
 		return fmt.Errorf("unable to obtain chain id: %w", err)
 	}
 
+	// if subzone
+	if p.SubzoneInfo != nil {
+		if chainID != p.SubzoneInfo.BaseChainID {
```
joe-bowman, line 34: If we don't store SubzoneInfo.BaseChainID separately (see previous comment) then we can remove this.

x/interchainstaking/keeper/proposal_handler.go
```diff
@@ -16,16 +16,37 @@ import (
...
+			return fmt.Errorf("incorrect ID \"%s\" for subzone \"%s\"", chainID, p.SubzoneInfo.BaseChainID)
+		}
+
+		// get zone
+		baseZone, found = k.GetZone(ctx, p.SubzoneInfo.BaseChainID)
+		if !found {
+			return fmt.Errorf("unable to find base chain \"%s\" for subzone \"%s\"", chainID, p.SubzoneInfo.BaseChainID)
+		}
+
```
joe-bowman, line 43: Assert parent zone is not a Subzone itself.

x/interchainstaking/keeper/proposal_handler.go
```diff
@@ -94,7 +124,7 @@ func (k *Keeper) HandleRegisterZoneProposal(ctx sdk.Context, p *types.RegisterZo
 
 	period := int64(k.GetParam(ctx, types.KeyValidatorSetInterval))
 	query := stakingTypes.QueryValidatorsRequest{}
-	err = k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ChainId, query, sdkmath.NewInt(period))
+	err = k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ID(), query, sdkmath.NewInt(period))
```
joe-bowman, line 127: `ChainID()`

x/interchainstaking/keeper/proposal_handler.go
```diff
@@ -239,32 +269,32 @@ func (k *Keeper) HandleUpdateZoneProposal(ctx sdk.Context, p *types.UpdateZonePr
 			k.SetZone(ctx, &zone)
 
 			// generate deposit account
-			portOwner := zone.ChainId + ".deposit"
+			portOwner := zone.ID() + ".deposit"
```
joe-bowman, line 272:
```suggestion
			portOwner := fmt.Sprintf("%s.%s", zone.ID(), icstypes.ICASuffixDeposit)
```

or maybe even `zone.DepositPortName()` helper functions?

x/interchainstaking/keeper/proposal_handler.go
```diff
@@ -239,32 +269,32 @@ func (k *Keeper) HandleUpdateZoneProposal(ctx sdk.Context, p *types.UpdateZonePr
...
-			portOwner = zone.ChainId + ".delegate"
+			portOwner = zone.ID() + ".delegate"
 			if err := k.registerInterchainAccount(ctx, zone.ConnectionId, portOwner); err != nil {
 				return err
 			}
 
 			period := int64(k.GetParam(ctx, types.KeyValidatorSetInterval))
 			query := stakingTypes.QueryValidatorsRequest{}
-			err := k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ChainId, query, sdkmath.NewInt(period))
+			err := k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ID(), query, sdkmath.NewInt(period))
```
joe-bowman, line 297: This does not need to happen for subzones.

x/interchainstaking/keeper/proposal_handler.go
joe-bowman: Actually, this does not need to happen for subzones.

x/interchainstaking/keeper/proposal_handler.go
joe-bowman: Given that this is already triggered on 127 and is a periodic ICQ, we only ever need to retrigger this if the `connection` was updated.

x/interchainstaking/keeper/validators.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // GetValidators returns validators by chainID.
 func (k Keeper) GetValidators(ctx sdk.Context, chainID string) []types.Validator {
-	validators := []types.Validator{}
```
joe-bowman, line 13: What is best practice here?

```
validators := []types.Validator{}
var validators []types.Validator
validators := make([]types.Validator, 0)
```

we should aim to be consistent, whatever the choice.

x/interchainstaking/keeper/zones.go
```diff
@@ -290,7 +290,7 @@ func (k *Keeper) SetAccountBalance(ctx sdk.Context, zone types.Zone, address str
 			k.ICQKeeper.MakeRequest(
 				ctx,
 				zone.ConnectionId,
-				zone.ChainId,
+				zone.ChainID(),
```
joe-bowman, line 293: `ID()`

x/interchainstaking/keeper/zones.go
```diff
@@ -308,7 +308,7 @@ func (k *Keeper) SetAccountBalance(ctx sdk.Context, zone types.Zone, address str
 		k.ICQKeeper.MakeRequest(
 			ctx,
 			zone.ConnectionId,
-			zone.ChainId,
+			zone.ChainID(),
```
joe-bowman, line 311: `ID()`

x/interchainstaking/keeper/zones.go
```diff
@@ -353,13 +353,13 @@ OUTER:
...
 	// send delegations to validators
-	k.Logger(ctx).Info("send performance delegations", "zone", zone.ChainId)
+	k.Logger(ctx).Info("send performance delegations", "zone", zone.ID())
 
 	msgs := make([]sdk.Msg, len(validatorsToDelegate))
 	for i, val := range validatorsToDelegate {
 		k.Logger(ctx).Info(
 			"performance delegation",
-			"zone", zone.ChainId,
+			"zone", zone.ID(),
```
joe-bowman, line 362: It's a log so not the endo of the world, but performance delegations will only happen on basechain `ChainID()`

x/interchainstaking/types/keys.go
```diff
@@ -119,7 +119,7 @@ func GetPerformanceDelegationKey(zone *Zone, delAddr sdk.AccAddress, valAddr sdk
 
 // GetPerformanceDelegationsKey gets the prefix for a delegator for all validators.
 func GetPerformanceDelegationsKey(zone *Zone, delAddr sdk.AccAddress) []byte {
-	return append(append(KeyPrefixPerformanceDelegation, []byte(zone.ChainId)...), delAddr.Bytes()...)
+	return append(append(KeyPrefixPerformanceDelegation, []byte(zone.ID())...), delAddr.Bytes()...)
```
joe-bowman, line 122: `ChainID()` because performance delegations.

x/interchainstaking/types/sub_zones.go
```diff
@@ -0,0 +1,93 @@
...
+	if z.IsSubzone() {
+		return z.SubzoneInfo.BaseChainID
+	}
+
+	return z.ChainId
+}
+
+// ID returns the unique identifier for the given zone.
+func (z *Zone) ID() string {
+	return z.ChainId
```
joe-bowman, line 27: Oh, I see what is happening now (re: my earlier question).

x/participationrewards/keeper/rewards_validatorSelection.go
```diff
@@ -17,7 +17,7 @@ import (
 // individually in a callback.
 func (k Keeper) AllocateValidatorSelectionRewards(ctx sdk.Context) {
 	k.icsKeeper.IterateZones(ctx, func(_ int64, zone *icstypes.Zone) (stop bool) {
```
joe-bowman, line 19: The PR score logic will only execute for baseChains. Might have to change how this is done, as currently it it computed on the fly for each zone - but in reality it need to be computed once for each baseChain and the scores reused for the subzone. This can wait until PR#2 though.

joe-bowman (changes requested):
Apologies; big ass review coming in. Still a bunch of ID vs ChainIDs to resolve, and question about ID vs ChainID vs BaseChainID.

x/interchainstaking/keeper/abci.go
```diff
@@ -56,7 +56,7 @@ func (k *Keeper) BeginBlocker(ctx sdk.Context) {
 						// trigger valset update.
 						period := int64(k.GetParam(ctx, types.KeyValidatorSetInterval))
 						query := stakingTypes.QueryValidatorsRequest{}
-						err := k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ChainId, query, sdkmath.NewInt(period))
+						err := k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ID(), query, sdkmath.NewInt(period))
```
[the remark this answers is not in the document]
aljo242: Should we just exclusively run the query for the base chain and skip for subzones?

x/interchainstaking/keeper/proposal_handler.go
aljo242: Yeah, I have these helper functions being added in https://github.com/ingenuity-build/quicksilver/pull/420.  Not sure if we should duplicate the work.

x/interchainstaking/keeper/proposal_handler.go
aljo242: We do this below in `ValidateSubzoneForBasezone()`

x/interchainstaking/keeper/validators.go
aljo242: i think the `var` declaration makes sense for 0 len slices you are building using `append()`.  And then only use `make()` when you are allocating something with actual `len` or `cap`

x/interchainstaking/keeper/abci.go
```diff
@@ -56,7 +56,7 @@ func (k *Keeper) BeginBlocker(ctx sdk.Context) {
 						// trigger valset update.
 						period := int64(k.GetParam(ctx, types.KeyValidatorSetInterval))
 						query := stakingTypes.QueryValidatorsRequest{}
-						err := k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ChainId, query, sdkmath.NewInt(period))
+						err := k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ChainID(), query, sdkmath.NewInt(period))
```
joe-bowman, line 59: this whole section from line 48 can be wrapped in `if !zone.IsSubzone()`

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -492,13 +494,18 @@ func (k *Keeper) HandleWithdrawForUser(ctx sdk.Context, zone *types.Zone, msg *b
 
 	period := int64(k.GetParam(ctx, types.KeyValidatorSetInterval))
 	query := stakingtypes.QueryValidatorsRequest{}
-	return k.EmitValSetQuery(ctx, zone.ConnectionId, zone.ChainId, query, sdkmath.NewInt(period))
+
+	if zone.IsSubzone() {
+		return nil
+	}
```
joe-bowman, line 500: Actually, we probably _do_ want to requery the valset here; because the withdrawal will have changed the state on the hostchain. Just needs directing at zone.ChainID()
```suggestion
```

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -624,7 +630,7 @@ func (k *Keeper) HandleBeginRedelegate(ctx sdk.Context, msg sdk.Msg, completion
 	k.ICQKeeper.MakeRequest(
 		ctx,
 		zone.ConnectionId,
-		zone.ChainId,
+		zone.ChainID(),
```
joe-bowman, line 633: I missed this one; it should be ID()

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -651,7 +657,7 @@ func (k *Keeper) HandleBeginRedelegate(ctx sdk.Context, msg sdk.Msg, completion
...
```
joe-bowman, line 660: Same here: ID()

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -734,7 +740,7 @@ func (k *Keeper) HandleUndelegate(ctx sdk.Context, msg sdk.Msg, completion time.
...
```
joe-bowman, line 743: ID()

x/interchainstaking/keeper/keeper.go
```diff
@@ -578,7 +578,7 @@ func (k *Keeper) GetAggregateIntentOrDefault(ctx sdk.Context, z *types.Zone) (ty
 	var filteredIntents types.ValidatorIntents
 
 	if len(z.AggregateIntent) == 0 {
-		intents = k.DefaultAggregateIntents(ctx, z.ChainId)
+		intents = k.DefaultAggregateIntents(ctx, z.ID())
```
joe-bowman, line 581: Assuming this iterates over all validators and gives them an equal distribution, I expect this ought to be ChainID()

joe-bowman (changes requested):
another couple of ID / ChainID we missed

joe-bowman:
Conflicts, ser

aljo242:
> Conflicts, ser

fixed!

joe-bowman (approved):
lgtm

PR #465, opened 2023-06-19
test: adding an endpoint to get withdrawal records for a user

muku314115:
## 1. Summary
Fixes # (issue)

## 2.Type of change

## 5. Checklist


proto/quicksilver/interchainstaking/v1/query.proto
```diff
@@ -170,6 +175,11 @@ message QueryWithdrawalRecordsResponse {
   cosmos.base.query.v1beta1.PageResponse pagination = 2;
 }
 
+message QueryUserWithdrawalRecordsRequest {
+  string user_address = 1 [(gogoproto.moretags) = "cosmos.AddressString"];
+  cosmos.base.query.v1beta1.PageRequest pagination = 3;
```
joe-bowman, line 180: 3?

muku314115: Sorry must have missed this, fixed it now

proto/quicksilver/interchainquery/v1/query.proto
```diff
@@ -16,7 +16,8 @@ option go_package = "github.com/ingenuity-build/quicksilver/x/interchainquery/ty
 service QuerySrvr {
   // Params returns the total set of minting parameters.
   rpc Queries(QueryRequestsRequest) returns (QueryRequestsResponse) {
-    option (google.api.http).get = "/quicksilver/interchainquery/v1/queries/{chain_id}";
+    option (google.api.http).get =
```
joe-bowman, line 19: @aljo242 - not sure why but something somewhere is causing Mukund and I to format proto in one way, and you another way, so we are constantly changing these lines back and forth between the two. Would be good to understand why this is happening!

joe-bowman (approved):
lgtm

proto/quicksilver/interchainquery/v1/query.proto
aljo242: @joe-bowman @muku314115 what is your `buf` version on your local machine?

aljo242: yeah i would move to latest which is `1.21.0`

I think that is causing the issue

ajansari95 (approved):
lgtm

aljo242 (approved):

PR #466, opened 2023-06-19
test: end point to get intents across all zones for a delegator address

muku314115:
## 1. Summary
Fixes # (issue)

## 2.Type of change

## 5. Checklist


x/interchainstaking/keeper/grpc_query_test.go
```diff
@@ -666,7 +794,7 @@ func (suite *KeeperTestSuite) TestKeeper_UserWithdrawalRecords() {
 				)
 			},
 			&types.QueryUserWithdrawalRecordsRequest{
-				UserAddress: testAddress,
+				UserAddress: "quick16pxh2v4hr28h2gkntgfk8qgh47pfmjfhzgeure",
```
joe-bowman, line 797: should this be hardcoded?

proto/quicksilver/interchainstaking/v1/query.proto
```diff
@@ -143,6 +151,14 @@ message QueryDelegatorIntentResponse {
   DelegatorIntent intent = 1;
 }
 
+message QueryDelegatorIntentsRequest {
+  string delegator_address = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+}
+
+message QueryDelegatorIntentsResponse {
+  repeated DelegatorIntent intents = 1;
```
joe-bowman, line 159: Does the DelegatorIntent contain the chain Id?

aljo242: It does not.

We should use something like: 

```proto
message DelegatorIntentsResponse {
     string chainID = 1;
     DelegatorIntent intent = 2;
}

message QueryDelegatorIntentsResponse {
  repeated DelegatorIntentsResponse intents = 1;
```

muku314115: added a chain id to the response as well with each intent

muku314115:
> @muku314115 is the target branch correct?

Built this on top of the Withdrawal records PR, so set that as the target branch

https://github.com/ingenuity-build/quicksilver/pull/465 this would come as the first one

PR #469, opened 2023-06-20
test: endpoint to get the receipt for a txHash

muku314115:
## 1. Summary
Fixes # (issue)

## 2.Type of change

## 5. Checklist


aljo242 (dismissed):

x/interchainstaking/keeper/grpc_query.go
```diff
@@ -193,6 +193,35 @@ func (k *Keeper) Receipts(c context.Context, req *types.QueryReceiptsRequest) (*
...
+	ctx := sdk.UnwrapSDKContext(c)
+
+	zone, found := k.GetZone(ctx, req.GetChainId())
+	if !found {
+		return nil, status.Error(codes.NotFound, fmt.Sprintf("no zone found matching %s", req.GetChainId()))
+	}
+
+	txReceipt := &types.Receipt{}
+
+	k.IterateZoneReceipts(ctx, &zone, func(_ int64, receipt types.Receipt) (stop bool) {
```
joe-bowman, line 214: No need to iterate the zones here:

`receipt, found := k.GetReceipt(types.GetReceiptKey(req.GetChainId(), req.GetTxHash()))`

joe-bowman: Strictly speaking we don't actually need to validate the zone, just the getReceipt would suffice - I guess the only reason to validate the zone is because is to tell the user they are looking for a chain that is not onboarded (e.g. a typo), but given that this endpoint is primarily used by the F/E I'm less inclined to care for this scenario.

joe-bowman (changes requested):
use `GetReceipt()` over iterating all receipts; as this will get slower over time.

x/interchainstaking/keeper/grpc_query.go
muku314115: added the changes and removed zone validation, returning receipt not found error in case of mismatch

aljo242 (commented):
lgtm after conflicts resolved

joe-bowman (commented):
lgtm, when conflicts are fixed.

ajansari95 (commented):
lgtm after conflicts are fixed

joe-bowman (approved):

ajansari95 (approved):

PR #472, opened 2023-06-21
chore: add `format` to `make proto-gen`

aljo242:
- Make sure `format` is run every time`make proto-gen` is run.
- run `make proto-gen`


ajansari95 (approved):

joe-bowman (approved):
lgtm

PR #475, opened 2023-06-27
Qck 501 intents as memofield

joe-bowman:
## 1. Summary
Fixes QCK-501

Legacy memoized intent with new memofields appended fails when validator address contains 0xFF 

Change makes intents a memorized field.

## 2.Type of change

- [x] This change requires that the relevant front end changes are deployed at the same time, as legacy memo format will no longer be supported. 

## 3. Implementation details

Memoized intent becomes a length-prefixed memofield, with prefix 0x02, removing the need for a separator (as addresses can contain 0x00 through 0xff so no separator character would be valid).

## 7. Future Work (optional)

Front end must be updated to prefix the memoized intent with two bytes - `[0x02, len(intent_bytes)]`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20-alpine3.17 AS builder
+FROM golang:1.20-alpine3.18 AS builder
 RUN apk add --no-cache git musl-dev openssl-dev linux-headers ca-certificates build-base
 
 WORKDIR /src/app/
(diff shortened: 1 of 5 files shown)


aljo242 (approved):
lgtm

This is a great simplification and improvement of what we already have !

ajansari95 (approved):
lgtm🚢

PR #477, opened 2023-06-28
fix: replace sdk.MustAddressFromBech32 with addressutils.AddressFromBech32 [QCK-502]

joe-bowman:
## 1. Summary
Fixes QCK-502

Replace uses of `sdk.MustAddressFromBech32` with the `addressutils` variant that does not require addresses to start with the local chain's HRP.

## 2.Type of change

### DIFF
--- a/x/participationrewards/keeper/rewards_holdings.go
+++ b/x/participationrewards/keeper/rewards_holdings.go
@@ -5,6 +5,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/ingenuity-build/quicksilver/utils"
+	"github.com/ingenuity-build/quicksilver/utils/addressutils"
 	airdroptypes "github.com/ingenuity-build/quicksilver/x/airdrop/types"
 	cmtypes "github.com/ingenuity-build/quicksilver/x/claimsmanager/types"
 	icstypes "github.com/ingenuity-build/quicksilver/x/interchainstaking/types"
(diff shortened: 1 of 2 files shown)


x/participationrewards/keeper/rewards_holdings.go
```diff
@@ -94,7 +95,7 @@ func (k Keeper) CalcUserHoldingsAllocations(ctx sdk.Context, zone *icstypes.Zone
 	tokensPerAsset := sdk.NewDecFromInt(zoneAllocation).Quo(sdk.NewDecFromInt(supply.Amount))
 
 	// determine ics rewards to be distributed per token.
-	icsRewardsAddr := sdk.MustAccAddressFromBech32(zone.WithdrawalAddress.Address)
+	icsRewardsAddr, _ := addressutils.AddressFromBech32(zone.WithdrawalAddress.Address, "")
```
ajansari95, line 98: Correct me if i am wrong, this will still use "quick" as prefix as HRP is empty

joe-bowman: no, "" means that no check will take place.

joe-bowman: We don't actually care for the prefix here. 

Although we should probably panic if the bech32 fails to decode, else we'll try to send from nil address and that won't behave well.

ajansari95: so just err check and looks good to me

x/participationrewards/keeper/rewards_holdings.go
```diff
@@ -52,6 +52,8 @@ func (k Keeper) CalcUserHoldingsAllocations(ctx sdk.Context, zone *icstypes.Zone
 
 	userAllocations := make([]types.UserAllocation, 0)
 	icsRewardsAllocations := make([]types.UserAllocation, 0)
+	icsRewardsBalance := sdk.NewCoins()
+	icsRewardsPerAsset := make(map[string]sdk.Dec, 0)
```
ajansari95, line 56: did we decide we will use var declarations, going forward? @aljo242  @joe-bowman

x/participationrewards/keeper/rewards_holdings.go
joe-bowman: I've updated to use the HRP check (as we know it should always be the `zone.AccountPrefix`) and handling the errors appropriately.

We also don't attempt to do this if zone.WithdrawalAddress has not been set yet.

ajansari95: amazing works 👍

x/participationrewards/keeper/rewards_holdings.go
joe-bowman: lint fails with this:

`x/participationrewards/keeper/rewards_holdings.go:53:2  prealloc  Consider pre-allocating `userAllocations``

ajansari95: so  for arrays, slices and other data structures preallocation is the best practice otherwise there can be race condition

ajansari95 (approved):
lgtm

aljo242 (approved):

PR #484, opened 2023-07-04
fix: remove use of iota

joe-bowman:
## 1. Summary
Remove instances of iota.

This avoids the unintended changing of variables through refactor, which is not picked up by tests as values all change relative to each other and tests still pass as a result.

No functional change.


ajansari95 (approved):
lgtm

muku314115 (approved):

muku314115 (approved):

ajansari95 (approved):
lgtm

PR #488, opened 2023-07-06
quietly ignore nil validator updates, without erroring, to avoid blocking icq

joe-bowman:
## 1. Summary
Fixes bug where nil validator response is received (query was made for validator that does not exist, valid nil proof delivered) but validator cannot be deleted so bil is never valid from a cosmos PoV.
 
## 2.Type of change

## 3. Implementation details

On receipt of `nil` validatorResponse, log error and return nil to allow further messages to be processed.


muku314115 (approved):

ajansari95 (approved):
lgtm

PR #501, opened 2023-07-13
Parse memo before looking up hash

joe-bowman:
## 1. Summary
Fixes issue with completed unbondings where memo is not parseable after change in format.

## 2.Type of change

## 3. Implementation details

ParseMemo field to correctly remove unbond/ prefix.

### DIFF
--- a/x/interchainstaking/keeper/ibc_packet_handlers_test.go
+++ b/x/interchainstaking/keeper/ibc_packet_handlers_test.go
@@ -518,7 +518,7 @@ func (suite *KeeperTestSuite) TestHandleWithdrawForUser() {
 				}
 			},
 			message: banktypes.MsgSend{},
-			memo:    "7C8B95EEE82CB63771E02EBEB05E6A80076D70B2E0A1C457F1FD1A0EF2EA961D",
+			memo:    "unbondSend/7C8B95EEE82CB63771E02EBEB05E6A80076D70B2E0A1C457F1FD1A0EF2EA961D",
 			err:     true,
 		},
 		{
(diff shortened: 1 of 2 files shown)


x/interchainstaking/keeper/ibc_packet_handlers_test.go
```diff
@@ -625,14 +625,14 @@ func (suite *KeeperTestSuite) TestHandleWithdrawForUser() {
 			}
 
 			quicksilver.InterchainstakingKeeper.IterateZoneStatusWithdrawalRecords(ctx, zone.ChainId, icstypes.WithdrawStatusSend, func(idx int64, withdrawal icstypes.WithdrawalRecord) bool {
-				if withdrawal.Txhash == test.memo {
+				if withdrawal.Txhash == test.memo[11:] {
```
joe-bowman, line 628: this is hacky. can we do:

```
hash, err := types.ParseTxMsgMemo(test.memo, types.MsgTypeUnbondSend)
			suite.Require().NoError(err)
			quicksilver.InterchainstakingKeeper.IterateZoneStatusWithdrawalRecords(ctx, zone.ChainId, icstypes.WithdrawStatusSend, func(idx int64, withdrawal icstypes.WithdrawalRecord) bool {

				if withdrawal.Txhash == hash {
					suite.Require().Fail("unexpected withdrawal record; status should be Completed.")
				}
				return false
			})
```

instead?

joe-bowman (commented):
I can't approve this, because I opened, but your changes lgtm :)

ajansari95 (approved):
LGTM

aljo242 (approved):

PR #507, opened 2023-07-18
fix: track acknowledgements via a field

ajansari95:
## 1. Summary
Fixes # (issue)
- better way to track the acknowledged  `UnDelegateMsg`

## 2.Type of change

## 4. How to test/use

- cant think of a better way then writing e2e test cause all the mocked states are being tested
- mocking the asynch flow wont help in covering anything substantial

## 5. Checklist


x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -499,7 +499,7 @@ func (k *Keeper) GCCompletedRedelegations(ctx sdk.Context) error {
 
 func (k *Keeper) HandleMaturedUnbondings(ctx sdk.Context, zone *types.Zone) error {
 	k.IterateZoneStatusWithdrawalRecords(ctx, zone.ChainId, types.WithdrawStatusUnbond, func(idx int64, withdrawal types.WithdrawalRecord) bool {
-		if ctx.BlockTime().After(withdrawal.CompletionTime) && !withdrawal.CompletionTime.Equal(time.Time{}) { // completion date has passed.
+		if ctx.BlockTime().After(withdrawal.CompletionTime) && withdrawal.Acknowledged { // completion date has passed.
```
aljo242, line 502: @ajansari95 can we add a test for this condition?

joe-bowman:
Logic looks good. Test case as Al says, plus we will need an upgrade handler for any existing in process unbondings else they will never complete

aljo242 (approved):

joe-bowman (approved):
lgtm

PR #509, opened 2023-07-18
fix possible nil performance address panic in participation rewards

joe-bowman:
## 1. Summary
Fixes panic in participationrewards epoch hook where performance account has not been initialised.

### DIFF
--- a/x/participationrewards/keeper/rewards_validatorSelection.go
+++ b/x/participationrewards/keeper/rewards_validatorSelection.go
@@ -17,23 +17,25 @@ import (
 // individually in a callback.
 func (k Keeper) AllocateValidatorSelectionRewards(ctx sdk.Context) {
 	k.icsKeeper.IterateZones(ctx, func(_ int64, zone *icstypes.Zone) (stop bool) {
-		k.Logger(ctx).Info("zones", "chain_id", zone.ChainId, "performance address", zone.PerformanceAddress.Address)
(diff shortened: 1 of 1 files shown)


ajansari95 (approved):
lgtm , waiting for checks

x/participationrewards/keeper/rewards_validatorSelection.go
```diff
@@ -17,23 +17,25 @@ import (
...
+				"cosmos.distribution.v1beta1.Query/DelegationTotalRewards",
+				bz,
+				sdk.NewInt(-1),
+				types.ModuleName,
+				ValidatorSelectionRewardsCallbackID,
+				0,
+			)
+		}
 		return false
 	})
```
aljo242, line 40: What should we do for the case where the perf address is `nil`?  Anything to report or handle?

ajansari95: i dont think so, if there are no performance delegation we dont have the benchmark on distribution. so we skip it

joe-bowman: Correct. We can skip the zone if the perf account isn’t set. It’s just a race condition after zone creation, between epoch end and ICA initialisation.

muku314115 (approved):

aljo242 (approved):

PR #510, opened 2023-07-18
update intent to use defaults as fallback when vp has not claimed

joe-bowman:
## 1. Summary
Ensure given user does not exercise disproportionate power over aggregateIntent with small number of claims.


aljo242 (commented):
Can we add test for this? or refactor out and test new logic?

ajansari95 (approved):
lgtm

PR #511, opened 2023-07-21
chore: remove usage of `suite.Require()` in tests

aljo242:
replace `suite.Require().*` with `suite.*` since it is a bit cleaner and the `Require()` is totally unnecessary


ajansari95 (approved):
LGTM

muku314115 (approved):

PR #512, opened 2023-07-21
refactor: expected keepers

aljo242:
Refactor our keepers to use their own `types.*Keeper` instead of importing the types directly.  This should untangle some of our dependencies.


ajansari95 (approved):
LGTM

muku314115 (approved):

PR #516, opened 2023-07-26
fix: handle redelegation ack

ajansari95:
## 1. Summary
Fixes # (issue)
- ensure the redelegation msgs are unique
- handle for GC redelegations in case of succesful `MsgBeginRedelegate`

## 2.Type of change

## 4. How to test/use
- `TestRemoveDuplicates`
- `TestReceiveAckForBeginRedelegate`

## 5. Checklist

### DIFF
--- a/x/interchainstaking/keeper/keeper.go
+++ b/x/interchainstaking/keeper/keeper.go
@@ -497,7 +497,7 @@ func (k *Keeper) GetRatio(ctx sdk.Context, zone types.Zone, epochRewards math.In
 func (k *Keeper) Rebalance(ctx sdk.Context, zone types.Zone, epochNumber int64) error {
 	currentAllocations, currentSum := k.GetDelegationMap(ctx, &zone)
 	targetAllocations := zone.GetAggregateIntentOrDefault()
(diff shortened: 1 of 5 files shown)


x/interchainstaking/keeper/keeper.go
```diff
@@ -523,7 +523,23 @@ type RebalanceTarget struct {
...
+		key := fmt.Sprintf("%v-%s-%s", r.Amount.String(), r.Source, r.Target)
+		if !encountered[key] {
+			encountered[key] = true
+			result = append(result, r)
+		}
+	}
+	return result
+}
+
+func DetermineAllocationsForRebalancing(currentAllocations map[string]math.Int, currentSum math.Int, targetAllocations types.ValidatorIntents, existingRedelegations []types.RedelegationRecord) RebalanceTargets {
```
aljo242, line 542: Should this function just always return the rebalance targets with the duplicates removed? 

@joe-bowman @ajansari95

ajansari95: yes

joe-bowman: Yes, the logic is completely changed in v1.4, so this is just a workaround to make do until then.

joe-bowman (approved):
lgtm

aljo242 (approved):

PR #539, opened 2023-08-18
chore: run workflows for unstable branches

aljo242:
Run test workflows on pushes to  `unstable*` branches so that we can get regular CodeCov reports for them


ajansari95 (approved):
lgtm

PR #547, opened 2023-09-07
Lsm types 46

joe-bowman:
Adds LSM types on v46 as a stub. We don't require the underlying logic in v1.2, just ability to encode/decode types.


faddat:
Is this good to go for merge?

PR #552, opened 2023-09-11
chore: golangci-lint run ./... --fix

faddat:
This PR lints the repository automatically, but does not do any corrections fo manual linting issues.


joe-bowman (approved):
lgtm

faddat:
sounds good.  This is just the autofixes from the new settings, but since we're merging to develop, I am okay with it not passing for now.  Should be able to fix that no later than my tomorrow.

PR #566, opened 2023-09-15
New quicksilverjs using grpc-gateway

hoangdv2429:
## 1. Summary

This PR contains a massively updated quicksilver-js. 

## 2.Type of change

## 4. How to test/use

from README.md but using createGrpcGatewayClient instead of the one in the manual.

## 5. Checklist

- [ ] update README
- [ ] testing using jest/spec
- [ ] fix eslint/type errors.
- [ ] fix camelCase and snake_case compability.


faddat (approved):

anhductn2001 (approved):

chalabi2:
@hoangdv2429 can you provide a little more insight here
```
fix eslint/type errors.
fix camelCase and snake_case compability
```
maybe an example of those issues.

hoangdv2429:
what I mean by that is fixing telescope: 
1. there are some type errors in the library (mostly with Timestamp, could just change it to Date so Typescript won't complain about it). but if we fix it, we'll fix it in Telescope. It still work, but I just get annoyed by those red mark.
2. for example, when you're querying for delegationUnbonding the type of delegationUnbonding is delegateAddress and validatorAddress but the actual fm.renderURL is using delegate_address and validator_address. (this is only an example, could not 100% named like this)
All of those errors should be fix in telescope and then regen the library.

hoangdv2429:
Screenshot 2023-09-17 at 13 46 26
take this as an example.

PR #601, opened 2023-09-16
force use of protobufjs 6.11.4

hoangdv2429:


sontrinh16 (approved):

faddat:
@hoangdv2429 please leave a description on each PR.  

For this one, I'd use 

"This pr foces the use of protobufsjs 6.11.4"

and I'd like to an upstream PR solving the issue at its source

faddat (approved):

hoangdv2429:
@faddat sorry sir, this kind of like an urgen resolve. I'll be careful and add description next time. Also about the issue, even if it's fixed in their repo. It'll take Cosmos-kit one week before publishing a new version that use newest version of protobufjs so my temper solution is this.

PR #618, opened 2023-09-18
monorepo: add icq relayer to quicksilver repository

faddat:
## 1. Summary
add icq relayer to new repository

## 2.Type of change

Add the replayer that qs uses to qs

## 3. Implementation details

In the future we will make releases of both in a matching way.


joe-bowman (approved):
lgtm

PR #623, opened 2023-09-18
ensure go version used in Makefile, Dockerfile and proto-gen match go.mod

joe-bowman:
Ensure we use go1.21 everywhere, and not a combination of the two.


ThanhNhann (commented):
Can we update the `go` version in `go.mod` @joe-bowman ?

joe-bowman:
> Can we update the `go` version in `go.mod` @joe-bowman ?

It was already updated there - this PR is to update it elsewhere.

odeke-em:
@joe-bowman might this be quite premature requiring everyone to install and use Go1.21.1 which was just released less than 14 days ago? Or is this change predicated upon some sort of freshly released security fix? It takes a while for people to get Go updates: even I as a core developer still has Go1.20.X

joe-bowman:
closed on keeping go1.20 for now.

PR #626, opened 2023-09-19
Add check tombstoned

### LINKED ISSUE #608: x/interchainstaking: Keeper.SetValidatorForZone should also check for tombstoned validators after retrieving the validator from store
## Summary of Bug
If you read through the code for Keeper.SetValidatorForZone
https://github.com/quicksilver-zone/quicksilver/blob/c26b22e7cf2eebef24f70289f73552f54da807ea/x/interchainstaking/keeper/keeper.go#L291-L348

we only seem to be checking if a validator wasn't jailed, but I suggest that for found validators let's also explicitly check that they weren't tombstoned because even if it were jailed and could unjail itself, tombstoning is an infinite suspension time

## Suggestion
```go
} else {
   if val.Tombstoned {
       k.Logger(ctx).Error("Tombstoned validator found", "valoper", validator.OperatorAddress)
       return fmt.Errorf("%q on chainID: %q was found to already have been tombstoned", validator.OperatorAddress, zone.ChainID)
   }

   if !val.Jailed && validator.IsJailed() {
       ...
   }
   ...
}
```

/cc @elias-orijtech  

## Version

____

#### For Admin Use

- [x] Not duplicate issue

ThanhNhann:
## 1. Summary
Fixes #608 

## 2.Type of change

## 3. Implementation details

- Add checking tombstones validator


sontrinh16 (approved):

joe-bowman (changes requested):
I believe the state change we want to check here is whether val.Jailed != validator.Jailed or if bondStatus has changed between the two.

Whatever the comparison we use, we need to trigger `toQuery = true`, to kick off the provable query.

ThanhNhann:
Got it, I will move the check like @odeke-em's suggestion to complete it

joe-bowman:
What si the status of this @ThanhNhann ?

ThanhNhann:
> What si the status of this @ThanhNhann ?

 I think we can close this and move to https://github.com/quicksilver-zone/quicksilver/pull/642

PR #627, opened 2023-09-19
Check negative amount in `GetUnbondingAmount` func

### LINKED ISSUE #610: x/interchainstaking: Keeper.GetUnbondingAmount should firstly check that .Amount is positive regardless of assumptions
## Summary of Bug
This code https://github.com/quicksilver-zone/quicksilver/blob/c26b22e7cf2eebef24f70289f73552f54da807ea/x/interchainstaking/keeper/zones.go#L100-L103 iterates over all the status withdrawal records and for each adds to the cumulative amount without checking if the values are negative

## Expected Behaviour
Let's just add a guard that WithdrawRecords don't have negative amounts so
```go
func (k *Keeper) GetUnbondingAmount(ctx sdk.Context, zone *types.Zone) sdk.Coin {
	out := sdk.NewCoin(zone.BaseDenom, sdk.ZeroInt())
	k.IterateZoneStatusWithdrawalRecords(ctx, zone.ChainId, types.WithdrawStatusUnbond, func(index int64, wr types.WithdrawalRecord) (stop bool) {
                 amount := wr.Amount[0]
                 if !amount.IsNegative() {
		    out = out.Add(amount)
                 }
		return false
	})
	return out
}
```

## Version

/cc @elias-orijtech 
____

#### For Admin Use

- [x] Not duplicate issue

ThanhNhann:
## 1. Summary
Fixes #610 

## 2.Type of change

## 3. Implementation details
- Add checking `negative` case of `amount` `WithdrawalRecord` in `GetUnbondingAmount`

### DIFF
--- a/x/interchainstaking/keeper/zones.go
+++ b/x/interchainstaking/keeper/zones.go
@@ -98,7 +98,10 @@ func (k *Keeper) GetDelegatedAmount(ctx sdk.Context, zone *types.Zone) sdk.Coin 
 func (k *Keeper) GetUnbondingAmount(ctx sdk.Context, zone *types.Zone) sdk.Coin {
 	out := sdk.NewCoin(zone.BaseDenom, sdk.ZeroInt())
 	k.IterateZoneStatusWithdrawalRecords(ctx, zone.ChainId, types.WithdrawStatusUnbond, func(index int64, wr types.WithdrawalRecord) (stop bool) {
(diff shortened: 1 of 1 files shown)


sontrinh16 (approved):

joe-bowman (approved):
This should not be possible to be Negative, but no harm being defensive.

lgtm.

PR #636, opened 2023-09-20
test for PerfBalanceCallback

### LINKED ISSUE #615: x/interchainstaking/keeper: Keeper.DepositTxCallback and Keeper.PerfBalanceCallback are untested
## Summary of Bug
The code for DepositTxCallback is untested and needs comprehensive testing.

- [x] Keeper.DepositTxCallback
- [x] Keeper.PerfBalanceCallback
## Version

____

#### For Admin Use

- [x] Not duplicate issue

DongLieu:
## 1. Summary
Fixes #615

## 2.Type of change

## 5. Checklist


x/interchainstaking/keeper/callbacks_test.go
```diff
@@ -1606,6 +1606,34 @@ func (suite *KeeperTestSuite) TestDelegationAccountBalanceCallback() {
...
+		zone, _ := quicksilver.InterchainstakingKeeper.GetZone(ctx, suite.chainB.ChainID)
+		zone.DepositAddress.IncrementBalanceWaitgroup()
+		zone.WithdrawalAddress.IncrementBalanceWaitgroup()
+		quicksilver.InterchainstakingKeeper.SetZone(ctx, &zone)
+
+		response := sdk.NewCoin("qck", sdk.NewInt(10))
+		respbz, err := quicksilver.AppCodec().Marshal(&response)
+		suite.NoError(err)
+
+		for _, addr := range []string{zone.DepositAddress.Address, zone.WithdrawalAddress.Address} {
```
joe-bowman, line 1627: We should probably be passing in the performance address here. Interesting that it works regardless of what address you return a query for; but we probably should check in this fn that we are using the correct address, else the following function could still fail.

DongLieu:
Is that okay?

joe-bowman:
Ideally, for me, this test would not just check that nothing errors, but the outcomes of this callback:
1) has the performance account balance been updated to the value the callback passes in
2) does it trigger performance delegations to be triggered, for those validators for which they do not already exist
  - this requires setting creating performance delegations for some of the zone B valopers, and then ensuring that the correct delegation messages are sent (I don't think we have a nice way of ensuring this presently; it would be nice to hook into k.SubmitTx (or perhaps replace it in test cases), but this is a bit more work.
 
(1) should be the bare minimum in this PR; (2) is possibly a future piece, unless you want to tackle it now.

DongLieu:
Have I done your request correctly?

ThanhNhann:
> 2. does it trigger performance delegations to be triggered, for those validators for which they do not already exist
> 
> * this requires setting creating performance delegations for some of the zone B valopers, and then ensuring that the correct delegation messages are sent (I don't think we have a nice way of ensuring this presently; it would be nice to hook into k.SubmitTx (or perhaps replace it in test cases), but this is a bit more work.
> 
> (2) is possibly a future piece, unless you want to tackle it now.

@anhductn2001 we should make a plan for interchaintest for func like this

ThanhNhann:
This pr linked with #615 so I updated the summary, and please fix the conflict for this @DongLieu

DongLieu:
Ok, I fixed it @ThanhNhann

ThanhNhann (approved):
Lgtm! thank @DongLieu

anhductn2001 (approved):

PR #642, opened 2023-09-24
Add ICQ to get `signingInfo` incase validator was jailed

### LINKED ISSUE #640: Add `ICQ` to get info of `Validator`
## Summary

- We should add an `ICQ` to get info of `Validator` to look up the validator sign info when to check whether `Validator` was `Tombstone` or not

#### For Admin Use

- [x] Not duplicate issue

ThanhNhann:
## 1. Summary
Fixes #640 & #608

## 2.Type of change


x/interchainstaking/keeper/keeper.go
```diff
@@ -259,6 +259,14 @@ func (k *Keeper) SetValidatorForZone(ctx sdk.Context, zone *types.Zone, data []b
 		return err
 	}
 
+	// In case validator is jailed
+	if validator.IsJailed() {
```
joe-bowman, line 263: This should also check that the validator wasn’t previously jailed, so we don’t keep resending the same request

x/interchainstaking/keeper/keeper.go
```diff
@@ -537,6 +522,28 @@ func (k *Keeper) EmitDepositIntervalQuery(ctx sdk.Context, zone *types.Zone) {
...
+	}
+	k.ICQKeeper.MakeRequest(
+		ctx,
+		connectionID,
+		chainID,
+		queryType,
+		bz,
+		sdk.NewInt(-1),
+		types.ModuleName,
+		"validatorigninginfo",
```
joe-bowman, line 540: Typo?

x/interchainstaking/keeper/callbacks.go
```diff
@@ -253,6 +255,101 @@ func DepositIntervalCallback(k *Keeper, ctx sdk.Context, args []byte, query icqt
...
 
+func ValidatorSigningInfosCallback(k *Keeper, ctx sdk.Context, args []byte, query icqtypes.Query) error {
+	zone, found := k.GetZone(ctx, query.GetChainId())
+	if !found {
+		return fmt.Errorf("no registered zone for chain id: %s", zone.ChainId)
+	}
+
+	k.Logger(ctx).Debug("Validator signing info callback", "zone", zone.ChainId)
+
+	signingInfoRes := slashingtypes.QuerySigningInfoResponse{}
```
joe-bowman, line 266: If we use the response here, there is no proof and the result can be forged. We can only update state based upon kv Store lookups

x/interchainstaking/keeper/keeper.go
```diff
@@ -537,6 +522,28 @@ func (k *Keeper) EmitDepositIntervalQuery(ctx sdk.Context, zone *types.Zone) {
 	)
 }
 
+func (k *Keeper) QueryValidatorSigningInfo(ctx sdk.Context, connectionID, chainID string, validator stakingtypes.Validator) error {
+	queryType := fmt.Sprintf("/cosmos/slashing/v1beta1/signing_infos/%s", validator.GetOperator().String())
```
joe-bowman, line 526: I believe this param is consensus are not operator address iirc

joe-bowman: And instead of hitting the SigningInfosQuery endpoint, we want to query the kv entry directly, with key add generated here: https://github.com/cosmos/cosmos-sdk/blob/d6d929843bbd331b885467475bcb3050788e30ca/x/slashing/keeper/signing_info.go#L14

x/interchainstaking/keeper/keeper.go
ThanhNhann: I collected all the logic of the process jailed in [ValidatorSigningInfosCallback](https://github.com/quicksilver-zone/quicksilver/blob/d980e666d07ec0cc60ddd935b1f40566e941665a/x/interchainstaking/keeper/callbacks.go#L258) so I think we will not resend the same request

x/interchainstaking/keeper/keeper.go
ThanhNhann: Yes, I forgot to update it, thank @joe-bowman

x/interchainstaking/keeper/keeper.go
ThanhNhann: Because if the validator wasn’t previously jailed, we also check if it now was jailed or was tombstones to update info right?

ThanhNhann: The logic now will make the infinite loop when `validator` was jailed, Im checking now

x/interchainstaking/keeper/callbacks.go
```diff
@@ -253,6 +255,30 @@ func DepositIntervalCallback(k *Keeper, ctx sdk.Context, args []byte, query icqt
...
+	if len(args) == 0 {
+		return errors.New("attempted to unmarshal zero length byte slice (10)")
+	}
+	err := k.cdc.Unmarshal(args, &valSigningInfo)
+	if err != nil {
+		return err
+	}
+	if valSigningInfo.Tombstoned {
+		k.Logger(ctx).Error("Tombstoned validator found", "valoper", valSigningInfo.Address)
+		return fmt.Errorf("%q on chainID: %q was found to already have been tombstoned", valSigningInfo.Address, zone.ChainId)
```
joe-bowman, line 276: As discussed on the call, we need to not error here, but instead, update the validator record and return nil (finding that a validator is tombstoned is considered 'success' because we can update our records to reflect that).

Will almost certainly need to update the validator record to store the consensus key of the validator, in order to later lookup the validator my consensus key.

x/interchainstaking/keeper/callbacks.go
```diff
@@ -273,7 +273,16 @@ func SigningInfoCallback(k *Keeper, ctx sdk.Context, args []byte, query icqtypes
 	}
 	if valSigningInfo.Tombstoned {
 		k.Logger(ctx).Error("Tombstoned validator found", "valoper", valSigningInfo.Address)
-		return fmt.Errorf("%q on chainID: %q was found to already have been tombstoned", valSigningInfo.Address, zone.ChainId)
+
+		valAddrBytes, err := addressutils.ValAddressFromBech32(valSigningInfo.Address, zone.GetValoperPrefix())
```
joe-bowman, line 277: Is this the consensus address, not the valoper?

x/interchainstaking/keeper/callbacks.go
```diff
@@ -273,7 +273,16 @@ func SigningInfoCallback(k *Keeper, ctx sdk.Context, args []byte, query icqtypes
...
+		if err != nil {
+			return err
+		}
+		val, _ := k.GetValidator(ctx, zone.ChainId, valAddrBytes)
+		val.Tombstoned = true
+		k.SetValidator(ctx, zone.ChainId, val)
```
joe-bowman, line 283: Should check the err returned (or return it). Valid to fail the tx here

x/interchainstaking/keeper/keeper.go
```diff
@@ -357,6 +353,13 @@ func (k *Keeper) SetValidatorForZone(ctx sdk.Context, zone *types.Zone, data []b
 		}
 	}
 
+	// we set the validator in the steps above so dont need to check found here
+	val, _ = k.GetValidator(ctx, zone.ChainId, valAddrBytes)
+	// checking whether validator was tombstoned or not
+	if val.Jailed {
+		k.EmitSigningInfoQuery(ctx, zone.ConnectionId, zone.ChainId, validator)
```
joe-bowman, line 360: Handle err here

x/interchainstaking/keeper/keeper.go
```diff
@@ -294,12 +290,12 @@ func (k *Keeper) SetValidatorForZone(ctx sdk.Context, zone *types.Zone, data []b
 		}
 
 	} else {
-		if !val.Jailed && validator.IsJailed() {
-			// checking in case validator was tombstoned
-			if err := k.EmitSigningInfoQuery(ctx, zone.ConnectionId, zone.ChainId, validator); err != nil {
-				return err
-			}
```
joe-bowman: I see what you mean now. I don’t believe the condition in l298 can be true if val.Tombstoned is true. 

If tombstoned is true, val.Jailed is already true.

x/interchainstaking/keeper/keeper.go
```diff
@@ -357,6 +353,13 @@ func (k *Keeper) SetValidatorForZone(ctx sdk.Context, zone *types.Zone, data []b
 		}
 	}
 
+	// we set the validator in the steps above so dont need to check found here
+	val, _ = k.GetValidator(ctx, zone.ChainId, valAddrBytes)
```
joe-bowman, line 357: Why are we getting the validator again here?

x/interchainstaking/keeper/keeper.go
ThanhNhann: Looks like you're reviewing my old code, right?

joe-bowman: Nope, the comment pertains to the new code. Reviewing on phone and clicked on the wrong line for the comment. Comment is accurate though.

ThanhNhann: I have a line to check `val.Tombstoned` before [here](https://github.com/quicksilver-zone/quicksilver/blob/c33d55844041b0af4951dc76304274b1ceb7e8b3/x/interchainstaking/keeper/keeper.go#L293), so I believe the condition still good

joe-bowman: The condition is fine in and of itself, but I don’t see why the check is needed (and the previous logic was already working afaics).

As before, erroring here will cause the request to never be satisfied, and thus be repeated indefinitely (along with failing all icq responses in the same tx).

unless we encounter a case that cannot be valid, (tombstoned is a genuine valid state), we must always return nil.

ThanhNhann: I see we should update the return, but I don't really understand about `but I don’t see why the check is needed `, can you make a `change suggestion` for ez understand?

x/interchainstaking/keeper/callbacks.go
ThanhNhann: Yeah this is `consensus address`, I have a mistake here, and we have to make another `ICQ` to get the validator address from this `consensus address` 
cc: [here](https://github.com/cosmos/cosmos-sdk/blob/e65083d9e466d995772d4db6011f4a2c1e9b070b/x/staking/keeper/validator.go#L47)

x/interchainstaking/keeper/callbacks.go
```diff
@@ -253,6 +255,42 @@ func DepositIntervalCallback(k *Keeper, ctx sdk.Context, args []byte, query icqt
...
+		return err
+	}
+	if valSigningInfo.Tombstoned {
+		k.Logger(ctx).Error("Tombstoned validator found", "valoper", valSigningInfo.Address)
+
+		valAddrBytes, err := addressutils.ValAddressFromBech32(valSigningInfo.Address, zone.GetValoperPrefix())
+		if err != nil {
+			return err
+		}
+		val, _ := k.GetValidator(ctx, zone.ChainId, valAddrBytes)
```
joe-bowman, line 281: k.GetValidatorByConsAddr()  <-- need to add this fn

x/interchainstaking/keeper/keeper.go
```diff
@@ -347,6 +353,14 @@ func (k *Keeper) SetValidatorForZone(ctx sdk.Context, zone *types.Zone, data []b
 		}
 	}
 
+	// checking whether validator was tombstoned or not
+	if val.Jailed {
+		err := k.EmitSigningInfoQuery(ctx, zone.ConnectionId, zone.ChainId, validator)
```
joe-bowman, line 358: Store validator.ConsensusAddr against val (need to add new field).

joe-bowman: Instead of adding a new field, add a new mapping between consaddress and valoper address.

We only need to store this for vals at the point we trigger a tombstone lookup.

x/interchainstaking/keeper/keeper.go
```diff
@@ -347,6 +353,14 @@ func (k *Keeper) SetValidatorForZone(ctx sdk.Context, zone *types.Zone, data []b
 		}
 	}
 
+	// checking whether validator was tombstoned or not
+	if val.Jailed {
```
joe-bowman, line 357: This wants to happen in the conditional at line 298 (`if !val.Jailed && validator.Jailed`) in order to avoid repeated lookups every time the validator VP changes once jailed (in this case, `val.Jailed == true`, but ` !val.Jailed && validator.Jailed` will evaluate to `true` only on the first run.

joe-bowman: We will also need to call this in l275, if the _new_ validator is in a jailed state.

ThanhNhann:
I think it's good now @joe-bowman, I'm considering adding the case like `val.Jailed && validator.IsJailed()`

x/interchainstaking/keeper/keeper.go
```diff
@@ -309,6 +345,12 @@ func (k *Keeper) SetValidatorForZone(ctx sdk.Context, zone *types.Zone, data []b
 				return err
 			}
 		} else if val.Jailed && !validator.IsJailed() {
+			consAddr, err := validator.GetConsAddr()
+			if err != nil {
+				return err
+			}
+			k.DeleteValidatorAddrByConsAddr(ctx, zone.ChainId, consAddr)
+
```
joe-bowman, line 353: I don't think this is necessary. Once the validator exists, it will always exist, right? Not sure of the purpose of removing it here.

x/interchainstaking/keeper/validator_test.go
```diff
@@ -77,17 +77,14 @@ func (suite *KeeperTestSuite) TestStoreGetDeleteValidatorByConsAddr() {
...
+		newValidator := stakingtypes.Validator{
+			OperatorAddress: validator.String(),
 			ConsensusPubkey: pkAny,
 		}
+		consAddr, err := newValidator.GetConsAddr()
+		suite.NoError(err)
 
-		err = app.InterchainstakingKeeper.SetValidatorAddrByConsAddr(ctx, zone.ChainId, newValidator)
+		app.InterchainstakingKeeper.SetValidatorAddrByConsAddr(ctx, zone.ChainId, newValidator.OperatorAddress, consAddr)
 		suite.NoError(err)
```
joe-bowman, line 88: what val of err is being checked here?

joe-bowman: I see SetValidatorAddrByConsAddr is  no longer returns.
```suggestion
```

x/interchainstaking/keeper/callbacks.go
```diff
@@ -253,6 +255,62 @@ func DepositIntervalCallback(k *Keeper, ctx sdk.Context, args []byte, query icqt
 	return nil
 }
 
+func SigningInfoCallback(k *Keeper, ctx sdk.Context, args []byte, query icqtypes.Query) error {
```
joe-bowman, line 258: Could you add a test for this?

Normal keeperSuite setup, add a consensus address for validatorSet[0] and then pass a SigningInfo into the callback handler and assert the validator record is marked as tombstoned.

x/interchainstaking/keeper/keeper.go
joe-bowman: The only thing to be aware of (worth noting somewhere) is that if consensus key rotation is added to cosmos, we'll need to store consensus keys for all validators, and add a check here for old vs new.

For now it's not an issue though.

ThanhNhann: Missing this, I will update in the next commit

x/interchainstaking/keeper/callbacks.go
```diff
@@ -281,10 +281,12 @@ func SigningInfoCallback(k *Keeper, ctx sdk.Context, args []byte, query icqtypes
 			return fmt.Errorf("can not get validator address from consensus address: %s", valSigningInfo.Address)
 		}
 
+		fmt.Println("founded")
```
ThanhNhann, line 284: I think this is used for debugging right @phamminh0811

phamminh0811: Yes

anhductn2001 (approved):
LGTM!

PR #657, opened 2023-09-29
Test tokenize shares

neitdung:
## 1. Summary
Fixes #617 

## 2.Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## 5. Checklist

- [ ] Does the Readme need to be updated?

PR #658, opened 2023-09-29
Handle tokenize share

### LINKED ISSUE #654: Function HandleTokenizedShares can never be executed exactly
## Summary of Bug

Function HandleTokenizedShares can't be executed exactly because the condition below always returns false:

https://github.com/quicksilver-zone/quicksilver/blob/8fc6f208090cb3154fde917fbbeaad6b928421d4/x/interchainstaking/keeper/ibc_packet_handlers.go#L546-L547

The type of `sharesAmount` is `sdk.Coin` but type of `dist.Amount` is `uint64` so this condition always returns false. Then logic inside this function can't be run.

## Expected Behaviour

We should compare `sharesAmount.Amount` with `dist.Amount` then HandleTokenizedShares can be run with expected logic.

____

#### For Admin Use

- [ ] Not duplicate issue

neitdung:
## 1. Summary
Fixes #654 

## 2.Type of change

## 3. Implementation details

Add data to WithdrawalRecord before send MsgTokenizeShare

## 4. How to test/use

Test in PR #657

## 5. Checklist


ThanhNhann (approved):
Lgtm! Good job @neitdung

ThanhNhann:
When reviewing this pr I see sometimes we name the variable which may confuse like the example below:
https://github.com/quicksilver-zone/quicksilver/blob/543f819670378f1be207be43054514583c7de99d/x/interchainstaking/keeper/withdrawal_record.go#L29
It's like `[]*types.Distribution` is a list of `type.Distribution` so we should use `distributions []*types.Distribution` instead of `distribution []*types.Distribution`. Cc @joe-bowman

joe-bowman:
> When reviewing this pr I see sometimes we name the variable which may confuse like the example below:
> 
> https://github.com/quicksilver-zone/quicksilver/blob/543f819670378f1be207be43054514583c7de99d/x/interchainstaking/keeper/withdrawal_record.go#L29
> 
> 
> It's like `[]*types.Distribution` is a list of `type.Distribution` so we should use `distributions []*types.Distribution` instead of `distribution []*types.Distribution`. Cc @joe-bowman

Distribution is one of those annoying English words where the plural is non-obvious. To me as a native English speaker, Distribution makes sense here, but given each individual distribution (vs the collective distribution) is called Distribution, I'm happy to use distributions for the slice :)

ThanhNhann:
> > When reviewing this pr I see sometimes we name the variable which may confuse like the example below:
> > https://github.com/quicksilver-zone/quicksilver/blob/543f819670378f1be207be43054514583c7de99d/x/interchainstaking/keeper/withdrawal_record.go#L29
> > 
> > It's like `[]*types.Distribution` is a list of `type.Distribution` so we should use `distributions []*types.Distribution` instead of `distribution []*types.Distribution`. Cc @joe-bowman
> 
> Distribution is one of those annoying English words where the plural is non-obvious. To me as a native English speaker, Distribution makes sense here, but given each individual distribution (vs the collective distribution) is called Distribution, I'm happy to use distributions for the slice :)

Let's make a pr to update it @neitdung

ThanhNhann (approved):
Lgtm! thank @neitdung

ThanhNhann:
Hey @anhductn2001, I think we should make an interchaintest for `HandleTokenizedShares` function to check whether it works correctly. I created the issue here #751

anhductn2001:
Ok @ThanhNhann, I will discuss with @neitdung and do it.

anhductn2001 (approved):
lgtm!

Issue #617, opened 2023-09-17
x/interchainstaking/keeper: untested handlers HandleCompleteSend, GCCompletedRedelegations, HandleTokenizedShares, DistributeRewardsFromWithdrawAccount, TriggerRedemptionRate, HandleWithdrawRewards, GetValidatorForToken, HandleRedeemTokens, HandleFailedUnbondSend, HandleFailedBankSend


odeke-em:
## Summary of Bug
These handlers are untested:
- [x] Keeper.HandleCompleteSend
- [x] Keeper.GCCompletedRedelegations
- [ ] Keeper.HandleTokenizedShares
- [x] Keeper.DistributeRewardsFromWithdrawAccount
- [x] Keeper.TriggerRedemptionRate
- [ ] Keeper.HandleWithdrawRewards
- [x] Keeper.GetValidatorForToken
- [x] Keeper.HandleRedeemTokens
- [x] Keeper.HandleFailedUnbondSend
- [x] Keeper.HandleFailedBankSend
- [x] msgServer.GovReopenChannel
- [x] msgServer.GovCloseChannel
- [ ] Keeper.HandleUpdateZoneProposal
- [ ] Keeper.EnsureWithdrawalAddresses
- [ ] Keeper.RemoveZoneAndAssociatedRecords
- [x] Keeper.GetActiveValidators
- [x] Keeper.SendTokenIBC
- [ ] Keeper.DeleteRedelegationRecordByKey
- [x] Keeper.GCCompletedUnbondings
- [x] Keeper.DeleteRedelegationRecordByKey
- [ ] Keeper.HandleUpdateZoneProposal
- [x] Keeper.Zone
- [x] Keeper.IteratePortConnections
- [x] Keeper.GetChainIDFromContext
- [x] PerfBalance
- [x] #710

## Version
13bf586baffd07ca4a2a7f093212455fafb5df27

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

hoangdv2429:
@DongLieu will take care of TestPerfBalanceCallback

neitdung:
We need to close #654 before write test for  **Keeper.HandleTokenizedShares**

neitdung:
Hey @ThanhNhann and @joe-bowman, I think Keeper.EnsureWithdrawalAddresses should be tested by interchain test. Because this function just return error when SubmitTx, every validations just return nil.
[...]

PR #659, opened 2023-10-02
Test for HandleWithdrawRewards

phamminh0811:
## 1. Summary
Fixes #617 

## 2.Type of change

## 5. Checklist


joe-bowman (dismissed):
lgtm

x/interchainstaking/keeper/ibc_packet_handlers_test.go
```diff
@@ -771,6 +773,168 @@ func (suite *KeeperTestSuite) TestHandleWithdrawForUserLSM() {
...
+			msg: func(zone *icstypes.Zone) sdk.Msg {
+				return &distrtypes.MsgWithdrawDelegatorReward{
+					DelegatorAddress: user,
+					ValidatorAddress: val,
+				}
+			},
+			triggered: false,
+			err:       true,
+		},
+		// at this time, the waitgroup will be uint32_max so it's not gonna trigger redemption rate
```
ThanhNhann, line 815: Can you explain this?

x/interchainstaking/keeper/ibc_packet_handlers_test.go
```diff
@@ -771,6 +773,168 @@ func (suite *KeeperTestSuite) TestHandleWithdrawForUserLSM() {
...
+		{
+			name: "negative waitgroup",
+			setup: func(ctx sdk.Context, quicksilver *app.Quicksilver, zone *icstypes.Zone) {
+				zone.WithdrawalWaitgroup = 0
```
ThanhNhann, line 819: I see we don't have a check whether `zone.WithdrawalWaitgroup < 0` or not, I think we should add it maybe in case the line below cc @joe-bowman  https://github.com/quicksilver-zone/quicksilver/blob/543f819670378f1be207be43054514583c7de99d/x/interchainstaking/keeper/ibc_packet_handlers.go#L1132

x/interchainstaking/keeper/ibc_packet_handlers_test.go
```diff
@@ -771,6 +773,168 @@ func (suite *KeeperTestSuite) TestHandleWithdrawForUserLSM() {
...
+				}
+			},
+			triggered: true,
+			err:       false,
+		},
+		{
+			name: "valid case trigger redemption rate and without set zone",
+			setup: func(ctx sdk.Context, quicksilver *app.Quicksilver, zone *icstypes.Zone) {
+				zone.WithdrawalWaitgroup = 0
+				quicksilver.InterchainstakingKeeper.SetZone(ctx, zone)
```
ThanhNhann, line 879: The name of the test case is `name: "valid case trigger redemption rate and without set zone",` but I also see you set zone in line `879`, Is anything wrong here 👀 I think this case is for checking `DelegatorAddress` is `zone.PerformanceAddress.Address` right?

x/interchainstaking/keeper/ibc_packet_handlers_test.go
ThanhNhann: I see in case `zone.WithdrawalWaitgroup = 0`, then if we continue minus it in this test case, the value of `zone.WithdrawalWaitgroup` will be equal to `uint32_max`, how about checking this in our logic? cc @joe-bowman

x/interchainstaking/keeper/ibc_packet_handlers_test.go
phamminh0811: This case should exist when we've just registered zone, then the WithdrawalWaitgroup = 0. But in this test we can't setup like that cause the zone need to be get right after suite.setupTestZones(). Do you think we need to separate this case to another test?

x/interchainstaking/keeper/ibc_packet_handlers_test.go
ThanhNhann: Can you make a pr for this @phamminh0811

ThanhNhann (approved):
Lgtm!

PR #673, opened 2023-10-08
Test HandleUpdateZoneProposal

phamminh0811:
## 1. Summary
Fixes #617

## 2.Type of change

## 5. Checklist


joe-bowman:
I note you added a check() funciton to the test case struct, but it is not used;

For the valid proposals here, can we also add a check() function so we can assert the values we wanted to set are indeed set against the zone in the kv store?

Thanks

ThanhNhann (approved):
Lgtm!

PR #678, opened 2023-10-09
Test for RemoveZoneAndAssociatedRecords

DongLieu:
## 1. Summary
Fixes #617

## 2.Type of change

## 5. Checklist


joe-bowman:
Can we also check for validators / receipts / withdrawal records etc. by zone to ensure that everything that should have been cleaned up has been? thankjs,

DongLieu:
Cannot check:

- delegation records: because the zone has been deleted

- performance delegation records: because the zone has been deleted

- validators: In the [RemoveZoneAndAssociatedRecords](https://github.com/quicksilver-zone/quicksilver/blob/main/x/interchainstaking/keeper/zones.go#L436) function do not remove Validator. Set add validators to create related

Additional:
In [GetReceipt](https://github.com/quicksilver-zone/quicksilver/blob/main/x/interchainstaking/keeper/receipt.go#L320), I think we should pass by value: 'chainID' and 'TxHash' instead of the current 'key'

joe-bowman:
Validators should definitely be removed if a zone is removed. Can you make a fix for this, please? 

Validator storage changed in v1.4 and the deletion of validators for a given zone was almost certainly overlooked.

joe-bowman:
> * delegation records: because the zone has been deleted
> * performance delegation records: because the zone has been deleted

Also probably a good time to refactor:

k.IterateAllDelegations(ctx, zone, func(delegation types.Delegation) (stop bool) {
k.IterateAllPerformanceDelegations(ctx, zone, func(delegation types.Delegation) (stop bool) {
k.IterateZoneReceipts(ctx, zone, func(index int64, receiptInfo types.Receipt) (stop bool) {
				
to take chainID instead of zone (all three methods only make use of zone.ChainID internally anyway!) but allows us to assert the removal properly.

DongLieu:
Okay,
Do I create a new PR to do this, or do it in this PR?

DongLieu:
For checks delegation and performance delegation records, I need to wait for the pull #748 to be merged

DongLieu:
I need to merge this PR #758 first, please review it

DongLieu:
Ok, done. @ThanhNhann @joe-bowman

x/interchainstaking/keeper/delegation_test.go
```diff
@@ -634,7 +634,7 @@ func (suite *KeeperTestSuite) TestFlushOutstandingDelegations() {
 func (suite *KeeperTestSuite) TestPerformanceDelegation() {
 	quicksilver := suite.GetQuicksilverApp(suite.chainA)
 	ctx := suite.chainA.GetContext()
-	chainID := "quicksilver-1"
```
ThanhNhann, line 637: Why we should use `const` here?

anhductn2001 (approved):
Nice works @DongLieu !

joe-bowman (approved):
lgtm

PR #691, opened 2023-10-11
[Validator] Add `GetValidatorByConsAddr`

ThanhNhann:
## 1. Summary
Fixes #  reference to [this](https://github.com/quicksilver-zone/quicksilver/pull/642#discussion_r1339714065)

## 2.Type of change


x/interchainstaking/keeper/validators.go
```diff
@@ -92,3 +92,34 @@ func (k Keeper) IterateValidators(ctx sdk.Context, chainID string, fn func(index
...
+	if len(bz) == 0 {
+		return val, false
+	}
+
+	k.cdc.MustUnmarshal(bz, &val)
+	return val, true
+}
+
+// SetValidatorByConsAddr set validator by Consensus address.
+func (k Keeper) SetValidatorByConsAddr(ctx sdk.Context, chainID string, val types.Validator) error {
```
joe-bowman, line 110: Do we want to store the val object here? Or simply the valoper address?

Storing the val object, means we have to update the validator in two places for any subsequent change (plus we are storing all the data twice). Just storing a reference to the valoper address is enough to lookup the canonical validator record however.

ThanhNhann: Make sense, I will update it

sontrinh16 (approved):

anhductn2001 (approved):

PR #699, opened 2023-10-12
Test for SendIbcToken

phamminh0811:
## 1. Summary
Fixes #617

## 2.Type of change

## 5. Checklist


x/interchainstaking/keeper/receipt_test.go
```diff
@@ -314,3 +317,58 @@ func (suite *KeeperTestSuite) TestReceiptStore() {
...
+		quicksilver.TransferKeeper.SetDenomTrace(
+			ctx,
+			transfertypes.DenomTrace{
+				Path:      fmt.Sprintf("%s/%s", portID, channelID),
+				BaseDenom: sdk.DefaultBondDenom,
+			},
+		)
+
+		// Try to send the ibc token
+		err = quicksilver.InterchainstakingKeeper.SendTokenIBC(ctx, sender, receiver, &zone, ibcAmount)
```
ThanhNhann, line 371: I think we should check the change in the receiver's balance after `SendTokenIBC` success, and add one more test case for [this](https://github.com/quicksilver-zone/quicksilver/blob/a24dc7029618e046d66b2af2028925cca7ba6338/x/interchainstaking/keeper/receipt.go#L140)

phamminh0811: There is no way to check the balance after testcase, refered to [TestSendTransfer](https://github.com/cosmos/ibc-go/blob/830982adfb4448573f8898bbd07a73d06bff640a/modules/apps/transfer/keeper/relay_test.go#L21) (ibc-go test)

ThanhNhann: Ok, just need to add one more test case then we can merge it

phamminh0811: @ThanhNhann added, please check

joe-bowman: Is it worth checking the balance of the escrow module as a check?

```
amount := quicksilver.TransferKeeper.GetTotalEscrowForDenom(suite.chainA.GetContext(), coin.GetDenom())
suite.Require().Equal(expEscrowAmount, amount.Amount)
```

sontrinh16 (approved):

x/interchainstaking/keeper/receipt_test.go
joe-bowman: @phamminh0811 did the suggestion above not work?

phamminh0811: there is no GetTotalEscrowForDenom function in transferkeeper

ThanhNhann: Can you check [this](https://github.com/cosmos/ibc-go/blob/3d9f652d11cece3dbd9aa9c204d7ef6b43b111b3/modules/apps/transfer/keeper/keeper.go#L220) @phamminh0811

joe-bowman (approved):
lgtm

x/interchainstaking/keeper/receipt_test.go
phamminh0811: @ThanhNhann I think it is on ibc-go v8. So right now we don't have that function yet

Issue #708, opened 2023-10-17
Delegation plan allocates to inactive validators


joe-bowman:
## Summary of Bug

Validators that are outside of the active set, but have historically received delegations, appear to be receiving delegations.

## Expected Behaviour

Prior to signalling intent, delegations should only be made to validators within the active set. Jailed, tombstones and otherwise unelectable validators ought not to receive delegations from Quicksilver.

## Version

v1.2 and v1.4 (main)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #709, opened 2023-10-17
Add delegation plan test

joe-bowman:
## 1. Summary

Address #708; delegation planning should not delegate to valdiators who are outside of the active set. This causes issues in production whereby validators are being delegated to, on the basis of having previously received a delegation, even if they have since been jailed or otherwise dropped out of the set.

## 2.Type of change

## 3. Implementation details

1. Add test to identify delegations made based upon incoming deposits that are allocated to invalid validators.
2. Fix delegation planning logic.

## 4. How to test/use

1. First commit will add a test based upon current mainnet state, and will failed on allocation to invalid validators.
2. Subsequent commits will rectify the underlying behaviour.


joe-bowman:
This PR is R4R now team :)

ThanhNhann (approved):
Nice work @joe-bowman. Lgtm!

sontrinh16 (approved):

PR #719, opened 2023-10-23
Update check stateless ICAAccount

### LINKED ISSUE #700: BalanceWaitgroup can not smaller than 0
## Summary of Bug

`balance_waitgroup`'s type is [uint32](https://github.com/quicksilver-zone/quicksilver/blob/ba0007795089637c8ab2524fd2546f89c98a73fe/proto/quicksilver/interchainstaking/v1/interchainstaking.proto#L70) so that it's smallest value will be zero and we don't need to check the value below 0 but if we continue decrease `BalanceWaitgroup`'s when it equal to 0, `BalanceWaitgroup`'s value will = 4294967295 (max uint32)
https://github.com/quicksilver-zone/quicksilver/blob/ba0007795089637c8ab2524fd2546f89c98a73fe/x/interchainstaking/types/accounts.go#L39-L40

## Expected Behaviour

Discuss in case `BalanceWaitgroup` = 0 and we keep decreasing it and update the stateless

## Steps to Reproduce

____

#### For Admin Use

- [x] Not duplicate issue

ThanhNhann:
## 1. Summary
Fixes #700 

- Update stateless check for `DecrementBalanceWaitgroup` func

## 2.Type of change

## 5. Checklist

### DIFF
--- a/x/interchainstaking/types/accounts.go
+++ b/x/interchainstaking/types/accounts.go
@@ -36,7 +36,7 @@ func (a *ICAAccount) IncrementBalanceWaitgroup() {
 }
 
 func (a *ICAAccount) DecrementBalanceWaitgroup() error {
-	if a.BalanceWaitgroup <= 0 {
+	if a.BalanceWaitgroup == 0 {
 		return errors.New("unable to decrement the balance waitgroup below 0")
 	}
 	a.BalanceWaitgroup--


sontrinh16 (approved):

joe-bowman (approved):
this will have no functional change, but for the sake of correctness, lgtm.

PR #735, opened 2023-10-25
change pass by value for GetReceipt and DeleteReceipt func

DongLieu:
I think the input for GetReceipt func should be changed to 'chainID' and 'txHash'


joe-bowman (approved):
lgtm

anhductn2001 (approved):

PR #752, opened 2023-10-30
Ensure ValidatorOperatorAddress is a correct bech32 address

### LINKED ISSUE #609: x/interchainstaking: Keeper.EmitValidatorQuery should check that the ValidatorOperatorAddress is a correct bech32 address and also return an error
## Summary of Bug
If at all the wrong operator address is passed in mistakenly, EmitValidatorQuery should be able to reject it and not emit the query but currently it assumes the address will always be bech32 per
https://github.com/quicksilver-zone/quicksilver/blob/c26b22e7cf2eebef24f70289f73552f54da807ea/x/interchainstaking/keeper/keeper.go#L500-L501

## Suggestion
Let's modify that signature of that method and then also check for errors returned
```go
_, addr, err := bech32.DecodeAndConvert(validator.OperatorAddress)
if err != nil {
   return fmt.Errorf("EmitValidatorQuery failed to decode validator.OperatorAddress: %q got error: %w",
         validator.OperatorAddress, err)
}
````

## Version

/cc @elias-orijtech 

____

#### For Admin Use

- [x] Not duplicate issue

### PR DESCRIPTION
**Pull request by neitdung**
## 1. Summary
Fixes #609 

## 2.Type of change

## 5. Checklist

### CODE REVIEW

**Comment by ThanhNhann** [comment]
Plz check the error of CI @neitdung

**Comment by joe-bowman** [comment]
This change feels superfluous? The address is already checked on keeper.go:219, but now we check it again and subsequently ignore the error in the only place it is used (keeper.go:241)? What are we trying to achieve here?

**Comment by neitdung** [comment]
Hi @joe-bowman, like your comment at here: [issue 609](https://github.com/quicksilver-zone/quicksilver/issues/609#issuecomment-1725097584). We do this for reuse in the future and puts the onus on the developer as it stands to ensure the correct error checking is done higher up.

File: x/interchainstaking/keeper/keeper.go [hunk 1]
```diff
@@ -238,7 +238,7 @@ func (k *Keeper) SetValidatorsForZone(ctx sdk.Context, data []byte, icqQuery icq
 		}
 
 		if toQuery {
-			k.EmitValidatorQuery(ctx, icqQuery.ConnectionId, icqQuery.ChainId, validator)
+			_ = k.EmitValidatorQuery(ctx, icqQuery.ConnectionId, icqQuery.ChainId, validator)
```
ThanhNhann, line 241: I think we should add a log here in case err

ThanhNhann (approved):
lgtm!

PR #753, opened 2023-10-30
Early return func AfterEpochEnd

### LINKED ISSUE #613: style: all: invert code and use early returns to make for direct reading without strained context and to reduce nesting
## Summary of Bug
Early returns are more idiomatic Go code and reduce the need for heavy nesting
If we look at the code all around the repository we see code nesting for example:
* Keeper.AfterEpochEnd could invert the conditions for an early return https://github.com/quicksilver-zone/quicksilver/blob/c26b22e7cf2eebef24f70289f73552f54da807ea/x/interchainstaking/keeper/hooks.go#L31-L196

can be inverted to
```go
if epochIdentifier != epochstypes.EpochIdentifierEpoch { // not at the Epoch's end!
    return nil
}

k.Logger(ctx).Info("handling epoch end", "epoch_identifier", epochIdentifier, "epoch_number", epochNumber)
```

We should do the same fixes throughout the entire repository.

## Version

____

#### For Admin Use

- [x] Not duplicate issue

neitdung:
## 1. Summary
Fixes #613 

## 2.Type of change

## 5. Checklist


ThanhNhann (commented):
Can you check another func has the same style like this

neitdung:
Ok, let me do this

neitdung:
Hi @ThanhNhann, I've read all module in `x/` folder and found 2 more issues. Please review this again.

x/mint/keeper/hooks.go
```diff
@@ -18,56 +18,57 @@ func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumb
 	params := k.GetParams(ctx)
```
ThanhNhann, line 18: I think we should move this line to after line 32 `minter := k.GetMinter(ctx)`

ThanhNhann (approved):
Lgtm!

PR #756, opened 2023-10-30
v1.2.17

joe-bowman:
When flushing delegations, we should ensure we can still calculate the redemption rate even if the delegation account balance is non-zero (e.g. if the MsgDelegate fails)

We do this by increasing the withdrawalWaitgroup by 1 for each MsgDelegate we send, and decrement the waitgroup on acknowledgement of the MsgDelegate execution (even if it failed). This way, we are able to use the withdrawalWaitgroup as the indicator that we have completed the cycle of delegation flushes, and rewards withdrawals, without having to rely on the delegation account balance being zero, which can legitimately not be the case.

Additionally, bump SDK to v0.46.15 and ibc-go to 5.3.2, and go to 1.20.


x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -189,7 +189,10 @@ func (k *Keeper) HandleAcknowledgement(ctx sdk.Context, packet channeltypes.Pack
 		case "/cosmos.staking.v1beta1.MsgDelegate":
 			// TODO: can we safely ignore this?
 			if !success {
-				return nil
+				if err := k.HandleFailedDelegate(ctx, src, packetData.Memo); err != nil {
+					return err
+				}
+				continue
```
joe-bowman, line 195: Note, we now `continue`, rather than `return nil` as we previously were able to assume that all `MsgDelegate` in a tx had failed, and no further action was required - but because we now decrement the waitgroup, we need to iterate through all failed messages.

ThanhNhann (approved):
Lgtm!

anhductn2001 (approved):
It's make sense. Nice works @joe-bowman

PR #766, opened 2023-11-02
fix: icq-relayer v0.10.0; add tm0.37 compat

joe-bowman:
## 1. Summary
- Fixes tm/comet v0.37 incompatibility in ICQ relayer
- Bump versions
- Changes module path based on monorepo inclusion

## 2.Type of change

## 3. Implementation details

This change replaces the call to RPCClient.Tx() with a local function that does not attempt to unmarshal the response, which is non-backward compatible between v0.34 and v0.37 (events are no longer base64 encoded, and are thusly stored as string instead of bytes).

The local function unmarshals into an struct containing only the proof and height fields (those we care about), and thus does not attempt to unmarshal the TxEvents.

## 4. How to test/use

Tested against Juno (tm0.37) and Txs now pass.


ThanhNhann (approved):
Lgtm!

sontrinh16 (approved):

PR #779, opened 2023-11-08
x/participationrewards/keeper: move cheap prefix checks to front + better error UX

### LINKED ISSUE #776: x/participationrewards: UmeeTotalBorrowsUpdateCallback could swap order of validation steps then better user experience to debug in case of mismatched prefixes
## Summary of Bug
Let's look at this code https://github.com/quicksilver-zone/quicksilver/blob/963ffa95f9712ab04df6a1f76ab8157e4d217620/x/participationrewards/keeper/callbacks.go#L207-L214. In there we can swap out the order of the first 2 calls and then also add context of the mismatched prefix vs it is expected.
image

/cc @elias-orijtech 

#### For Admin Use

- [x] Not duplicate issue

odeke-em:
This change moves the cheaper prefix checks per request to the top of each of UmeeTotalBorrowsUpdateCallback and UmeeInterestScalarUpdateCallback, prefix checks as it is unnecessary to invoke an expensive unmarshalling yet the result will be discarded from the result of a simple equals check. While here, also improved the error user experience (UX) for prefix mismatches to return the prefix that was retrieved vs expected.

Fixes #776

### DIFF
--- a/x/participationrewards/keeper/callbacks.go
+++ b/x/participationrewards/keeper/callbacks.go
@@ -204,13 +204,13 @@ func UmeeReservesUpdateCallback(ctx sdk.Context, k *Keeper, response []byte, que
 }
 
 func UmeeTotalBorrowsUpdateCallback(ctx sdk.Context, k *Keeper, response []byte, query icqtypes.Query) error {
+	if g, w := query.Request[0], umeetypes.KeyPrefixAdjustedTotalBorrow[0]; g != w {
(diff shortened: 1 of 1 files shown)


ThanhNhann:
hey @anhductn2001, can you plz check the CI, so we should extend the time for testing now

ThanhNhann (commented):
Lgtm! Just need to wait for @anhductn2001 to update the CI then we can merge

ThanhNhann (approved):
Lgtm! Thank @odeke-em

PR #789, opened 2023-11-18
defensive checks in UpdateWithdrawalRecordsForSlash

### LINKED ISSUE #607: x/interchainstaking: Keeper.UpdateWithdrawalRecordsForSlash should validate that each Distribution.Valoper's amount is positive before additions otherwise an overflow can trigger wrong results
## Summary of Bug
This code https://github.com/quicksilver-zone/quicksilver/blob/bbf9ef90b82b0afde218e146d5c3c94da867c4dd/x/interchainstaking/keeper/keeper.go#L362-L363 blindly creates an `sdk.NewInt(int64(d.Amount))` then performs operations on it assuming the result will always be positive! Same as in https://github.com/quicksilver-zone/quicksilver/blob/bbf9ef90b82b0afde218e146d5c3c94da867c4dd/x/interchainstaking/keeper/keeper.go#L369

Firstly let's note that `WithdrawalRecord.Distribution[].Amount` is a uint64 per https://github.com/quicksilver-zone/quicksilver/blob/c26b22e7cf2eebef24f70289f73552f54da807ea/proto/quicksilver/interchainstaking/v1/interchainstaking.proto#L86

If someone wanted to control how much money was to be withdrawn they could intercept traffic or maliciously set a large to be sent into amount that's larger than int64 (given .Amount is a uint64) and that would cause a negative value which could then when subtracted from record.Amount cause a large value or any value of the attacker's choosing

## Suggestion
```go
dAmount := sdk.NewDec(int64(d.Amount))
if dAmount.IsNegative() {
    // TODO: Log or report or punish for this serious overflow
    continue
}
```

then also for
https://github.com/quicksilver-zone/quicksilver/blob/bbf9ef90b82b0afde218e146d5c3c94da867c4dd/x/interchainstaking/keeper/keeper.go#L369 we could do
```go
deductedTotal := record.Amount.Sub(sdk.NewCoin(zone.BaseDenom, recordSubAmount))
if deductedTotal.GTE(record.Amount) {
    // TODO: Log or report or punish for this serious overflow
    return false
}

record.Distribution = distr
record.Amount = deductedTotal
k.SetWithdrawalRecord(ctx, record)
```

## Version

/cc @elias-orijtech 
____

#### For Admin Use

- [x] Not duplicate issue

joe-bowman:
## 1. Summary
Fixes #607 

- Add defensive checks against overflows
- Add tests
- Move UpdateWithdrawalRecordsForSlash to withdrawal_record.go

## 2.Type of change

## 3. Implementation details

Throw an error on if slashing will cause distAmount or record.Amount to be negative.


faddat (approved):
lgtm

PR #847, opened 2023-12-10
[chore] update golangci to v1.55.2

faddat:
## 1. Summary
Update golangci and fix some issues. 

Earlier, this PR used "latest" but it turns out you just need to specify the patch version when choosing a golangci-lint version :).


faddat:
@joe-bowman I think this is good to go but can you give it a once-over? don't want to ninja merge this since it has actual changes.

joe-bowman (approved):
ship it

PR #868, opened 2023-12-11
Update LICENSE to mit with riders

faddat:
## 1. Summary

Makes governance the only licensor of the code.


LICENSE
```diff
@@ -1,201 +1,9 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+Copyright <YEAR> <COPYRIGHT HOLDER>
```
joe-bowman, line 1: Does this need completing?

faddat: sure does

faddat:
PS: does this look right to you?  I am not 100% sure nor am I a lawyer.  But I love the idea of gov controlling the use of the chain code.

joe-bowman:
Do we need to ask governance to change this license in the first instance?

faddat:
I guess I don't think so?  

As the contributors / originaltors of the original license, I think we could just change it, HOWEVER!

I think we should ask governance.  :)

joe-bowman (approved):
Ship it

PR #875, opened 2023-12-11
Defensive checks callbacks

### LINKED ISSUE #842: x/participationrewards/keeper: Callbacks.Call needs bulletproofing to prevent nil pointer derefernce if no such callback is found
## Summary of Bug
Code that's in RPCs and especially callbacks are susceptible to arbitrary input from the outside world. For example please see this code

https://github.com/quicksilver-zone/quicksilver/blob/78e724f13bece830869cf71129d9912f02bc2e86/x/participationrewards/keeper/callbacks.go#L51-L53
which in one shot retrieves the callback function and then passes arguments to it. Sadly that code will panic if an unsaved id is passed in, because the code author might have presumed that by the time that code is called, the callback will always be set, or they were being clever or rushed, but to have secure code, please don't rely on assumptions like this.

We can protect against nil pointer dereference panics when a callback isn't found in the map simply by checking if a result wasn't found or not 
image

```go
cb, ok := c.Callbacks[id]
if !ok {
   return fmt.Errorf("no such callback %q", id)
}
return cb(ctx, c.id, args, ...)
```

/cc @faddat @joe-bowman @elias-orijtech 
____

#### For Admin Use

- [ ] Not duplicate issue

joe-bowman:
## 1. Summary
Fixes #842

Adds additional check to avoid nil pointer.

## 2.Type of change

### DIFF
--- a/utils/strings.go
+++ b/utils/strings.go
@@ -1,12 +1,5 @@
 package utils
 
-// LengthPrefixString returns length-prefixed bytes representation of a string.
-func LengthPrefixString(s string) []byte {
-	bz := []byte(s)
-	bzLen := len(bz)
-	return append([]byte{byte(bzLen)}, bz...)
-}
-
 // create a map from a slice of strings for efficient lookup
 func StringSliceToMap(in []string) map[string]bool {
 	out := make(map[string]bool, len(in))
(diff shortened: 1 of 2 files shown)


faddat (approved):

x/participationrewards/keeper/callbacks.go
```diff
@@ -49,6 +49,9 @@ func (k *Keeper) CallbackHandler() Callbacks {
 
 // Call calls callback handler.
 func (c Callbacks) Call(ctx sdk.Context, id string, args []byte, query icqtypes.Query) error {
+	if c.Has(id) {
```
joe-bowman, line 52: this should be negated. arse.

faddat: like !if ?

PR #913, opened 2023-12-19
remove assets page

faddat:
This PR removes the assets page from the quicksilver webui, enabling it to build, and enabling us to test its deployment to github pages.


faddat:
hmmmm will unfix my "fix"

faddat:
@chalabi2 - after you review and merge this, it should be published to github pages.  If it doesn't, lmk 🖖

faddat:
I'll also make an issue so that CI for the chain doesn't slow down web dev...

chalabi2 (approved):
Removes assets page with no breaking changes

chalabi2 (approved):

chalabi2:
> I'll also make an issue so that CI for the chain doesn't slow down web dev...

So long as the web-ui directory never depends on chain or js(npm packages only) stuff we should be okay.

faddat:
@chalabi2 the chain CI can take a long time to run, so I'm going to roll with it.  Need to get your approvals a little more power, but I think we are moving good :)!

PR #921, opened 2023-12-23
Delete web-ui/LISCENSE

faddat:
We are going to try and exotic license and that means that this license that was put here to stop others from copy pasta chalabi work, must die


joe-bowman (approved):
Lgtm

PR #922, opened 2023-12-23
Remove dependabout from Quicksilver.s and add depend about to web UI

faddat:
## 1. Summary
Just keeping dependabot happy


joe-bowman (approved):
Lgtm

PR #934, opened 2023-12-24
update .gitignore

faddat:
Remove webui gitignore and then update main gitignore


faddat:
Going on the ninja list due to not touching chain our ui code

PR #938, opened 2023-12-24
ensure we have standard 6 second blocks

faddat:
## 1. Summary

While two second blocks are great, we should not do that till we are on 50.

## 2.Type of change

## 5. Checklist


joe-bowman (approved):
lgtm

PR #939, opened 2023-12-24
fix issue in defensive checks

faddat:
## 1. Summary

Negate the check on line 52 so we can correctly see weather or not a chain is using epochs and what is the epoch block.

## 2.Type of change

## 5. Checklist


joe-bowman (approved):
Lgtm

PR #969, opened 2023-12-27
Release/v1.4.5-rc2

joe-bowman:
## 1. Summary
Pre v1.4.5 release fixes.

- Bump dependencies
- Lint fixes
- Make supply endpoint configurable
- Improve build tools


x/interchainstaking/keeper/redemptions.go
```diff
@@ -16,6 +14,7 @@ import (
 	"github.com/quicksilver-zone/quicksilver/utils"
 	epochstypes "github.com/quicksilver-zone/quicksilver/x/epochs/types"
 	"github.com/quicksilver-zone/quicksilver/x/interchainstaking/types"
+	lsmstakingtypes "github.com/quicksilver-zone/quicksilver/x/lsmtypes"
```
faddat, line 17: awesome

faddat (approved):
let's rock

joe-bowman:
@faddat - just added noop upgradehandler for testnet. please take another look ❤️

cmd/quicksilverd/root.go
```diff
@@ -134,8 +135,8 @@ func initTendermintConfig() *tmcfg.Config {
 	cfg := tmcfg.DefaultConfig()
 
 	// peers
-	cfg.P2P.MaxNumInboundPeers = 200
-	cfg.P2P.MaxNumOutboundPeers = 40
+	cfg.P2P.MaxNumInboundPeers = 30
```
faddat, line 138: generally would go higher and 8:1 ratio, but this works and will be much better than the defaults -- when it says it puts a strain on RAM it means ~8MB

x/interchainstaking/keeper/callbacks.go
```diff
@@ -297,6 +298,7 @@ func SigningInfoCallback(k *Keeper, ctx sdk.Context, args []byte, query icqtypes
 			return err
 		}
 		val, found := k.GetValidator(ctx, zone.ChainId, valAddrBytes)
+		// NOTE: this shouldn't be reachable, but keeping here as it doesn't do any harm.
```
faddat, line 301: 👍

x/interchainstaking/keeper/keeper.go
```diff
@@ -516,6 +511,9 @@ func (k *Keeper) EmitPerformanceBalanceQuery(ctx sdk.Context, zone *types.Zone)
 		100,
 	)
 
+	zone.PerformanceAddress.BalanceWaitgroup = 1
```
faddat, line 514: intending to use, or eliminate concurrency?

joe-bowman: Not sure what you mean?

faddat (approved):
looks good the second time, too!

only question is the waitgroup, if it's "1" to keep non determinism at bay, then let's do this

x/interchainstaking/keeper/keeper.go
faddat: balancewaitgroup?

this ins't a threading thing?

joe-bowman:
Ah - so the balanceWaitgroup logic on the ICAAccount interface is designed to support querying for multiple denoms - hence the waitgroup, which is used to determine when we have received the proofs for each denom.

For the performance account we only ever want to know about the baseDenom, so balance waitgroup can be zero or one.

It is decremented by the standard logic in the AccountBalance callback, but incrementing here, rather than setting to one can cause issues if there is a race condition between two calls to this method (the wg gets incremented twice, but the second icq request _replaces_ the first, so we'd only ever get decremented once. It's not strictly an issue here, because we don't ever rely on the wg returning to zero for this code path, but clean accounting is nice :)

PR #978, opened 2023-12-29
fix: permit zero rewards in validatorselectionrewardscallback

### LINKED ISSUE #977: ValidatorSelectionRewardsCallback expects non-zero rewards
## Summary of Bug

The ValidatorSelectionRewardsCallback in x/participation rewards expects non-zero rewards to have been earned in the last epoch, and emits an error if zero response is received. This is atypical, but possible, and should be acceptable.

## Expected Behaviour

Callback execution with zero rewards is valid and should be acceptable.

## Version

v14.5.-rc2

## Steps to Reproduce

Submit a ValidatorSelectionRewardsCallback with a payload containing zero rewards:

```
icq-provider_1  | ts=2023-12-29T15:51:46.635153648Z caller=run.go:767 worker=flusher chain=rhye-2 msg="Failed to submit; nevermind, we'll try again!" err="rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 4: no delegator rewards [cosmos/cosmos-sdk@v0.46.16/baseapp/baseapp.go:809] With gas wanted: '75000000' and gas used: '1218555' : unknown request"
```

____

#### For Admin Use

- [ ] Not duplicate issue

joe-bowman:
## 1. Summary
Fixes #977 

Tolerate payload containing zero accrued rewards by the performance account in the previous epoch.

Short circuit and return nil.

## 2.Type of change

## 3. Implementation details

Nil rewards are entirely valid (although uncommon), and we should accept this as a valid value.

### DIFF
--- a/x/participationrewards/keeper/rewards_validatorSelection.go
+++ b/x/participationrewards/keeper/rewards_validatorSelection.go
@@ -162,17 +162,13 @@ func (k Keeper) CalcOverallScores(
 
 	rewards := delegatorRewards.GetRewards()
 	if rewards == nil {
-		err := errors.New("no delegator rewards")
-		k.Logger(ctx).Error(err.Error())
-		return err
+		return nil
 	}
 
 	total := delegatorRewards.GetTotal().AmountOf(zone.BaseDenom)
 
 	if total.IsZero() {
-		err := errors.New("no delegator rewards (2)")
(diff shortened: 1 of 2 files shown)


faddat:
this is another great catch!

faddat (approved):

PR #983, opened 2023-12-31
tools: add upgrade handler for v1.4.5-rc4

joe-bowman:
## 1. Summary
Add's a no-op upgradehandler to facilitate the upgrade to v1.4.5-rc4 of the rhye-2 testnet.


faddat (approved):
good good

PR #989, opened 2024-01-02
fix: no need to allocate dust, given it might exceed a cap; fixed #988

### LINKED ISSUE #988: Allocating dust may breach LSM cap
## Summary of Bug

When allocating a deposit amongst validators on an LSM enabled chain, we need to ensure chain caps are not breached. Recent changes cater for this constraint well, except the final act is always to allocate the remaining dust (usually a few minor tokens; e.g. <10uatom) but this can potentially cause the lsm cap to be exceeded,as  we blindly allocate to the first validator in the list, that may or may not have capacity remaining.

## Expected Behaviour

We do not, in fact, need to allocate dust, as it will be picked up and delegated at the end of each epoch. We should ignore the dust portion, as it is expected to be tiny.

## Version

v1.4.5-rc4

joe-bowman:
## 1. Summary
Fixes 988

Remove the allocation of dust at the end of the delegation attempt.

## 2.Type of change

## 3. Implementation details

We currently attempt to allocation delegations amongst validators subject to delegation caps. At the end of the delegation planning, we are often left with some small amount of dust, and this has historically been allocated to the first validator in the list.

However, it is possible that this validator has reached it's cap, and thus the delegation will fail. As a result, _not_ delegating the dust is considered a safer option. Remaining dust will be delegated at the end of the epoch.


faddat (approved):
Dust bad.  Aggregation good.

PR #997, opened 2024-01-03
Disable lsm unbond

joe-bowman:
Remove LSM unbonding given it's unreliability, and always use native unbonding.


faddat:
Do you think this should result in any upstream fixes?

faddat (dismissed):
Great catch we love to live reliable

joe-bowman:
> Do you think this should result in any upstream fixes?

No, just LSM unbonding is a much more complex math problem, as the user receives bonded tokens back, and you need to compromise somewhere between what is best for the protocol, and what is best for the user (not receiving buckets of dust). It requires a little more finesse.

faddat (approved):
finesse

PR #1001, opened 2024-01-04
perf: avoid tiny delegations and redelegations

joe-bowman:
Improves efficiency: we sometimes find ourselves delegating or redelegating amounts that are smaller than the gas we spend doing the operation. Don't delegate or redelegate less than one major denom unit. 

Filter out msgs for delegations or redelegations < than 1000000 minor denom units. 

Future work: This should be configurable per zone (0. 1 atom is probably fine to delegate; 0.1 star less so).


faddat:
I would argue not more than ten, since we have the intents system in place.  

This would greatly reduce the number of ica transactions

x/interchainstaking/keeper/delegation.go
```diff
@@ -195,7 +195,10 @@ func (*Keeper) PrepareDelegationMessagesForCoins(zone *types.Zone, allocations m
 	var msgs []sdk.Msg
 	for _, valoper := range utils.Keys(allocations) {
 		if allocations[valoper].IsPositive() {
-			msgs = append(msgs, &stakingtypes.MsgDelegate{DelegatorAddress: zone.DelegationAddress.Address, ValidatorAddress: valoper, Amount: sdk.NewCoin(zone.BaseDenom, allocations[valoper])})
+			if allocations[valoper].GTE(sdk.NewInt(1_000_000)) {
```
joe-bowman, line 198:
```suggestion
			if allocations[valoper].GTE(sdk.NewInt(10_000_000)) {
```

x/interchainstaking/keeper/keeper.go
```diff
@@ -725,14 +725,17 @@ func (k *Keeper) Rebalance(ctx sdk.Context, zone *types.Zone, epochNumber int64)
...
 	for _, rebalance := range rebalances {
-		msgs = append(msgs, &stakingtypes.MsgBeginRedelegate{DelegatorAddress: zone.DelegationAddress.Address, ValidatorSrcAddress: rebalance.Source, ValidatorDstAddress: rebalance.Target, Amount: sdk.NewCoin(zone.BaseDenom, rebalance.Amount)})
-		k.SetRedelegationRecord(ctx, types.RedelegationRecord{
-			ChainId:     zone.ChainId,
-			EpochNumber: epochNumber,
-			Source:      rebalance.Source,
-			Destination: rebalance.Target,
-			Amount:      rebalance.Amount.Int64(),
-		})
+		if rebalance.Amount.GTE(sdk.NewInt(1_000_000)) {
```
joe-bowman, line 728:
```suggestion
		if rebalance.Amount.GTE(sdk.NewInt(10_000_000)) {
```

joe-bowman:
I'd rather leave this was 1m tokens for now, until we can configure one per zone. Will massively improve current situation, but 10m minor tokens (given this is per val) - could mean deposits of 18.5k USD (185 vals * 9.9 atoms * atom price) don't get deposited (in worst case) until the flush happens.

faddat (approved):

PR #1003, opened 2024-01-04
fix: we should still set intent when we have zero existing claims

### LINKED ISSUE #1002: Intents not being set when existing claims value is nil
## Summary of Bug

When the user does not have existing claims, the intent set by memo or lsm coins is not being set for the user.

## Expected Behaviour

The user's intent should be updated when staking regardless of whether the user has made a claim or not.

## Version

v1.4.5-rc6

joe-bowman:
## 1. Summary
Fixes #1002 - Ensure setIntent is called correctly even when the user has zero existing claims.

## 2.Type of change


faddat (approved):
1 deleted line of code

Dozens of tests.

Love it!

PR #1019, opened 2024-01-09
Release/v1.4.6

joe-bowman:
## 1. Summary
Fixes v1.4.5 failed upgrade.

## 2.Type of change

## 3. Implementation details

The upgrade handler for v1.4.5 release did not execute, as the registered handler reference a no-op handler (i.e. has not impact), rather than one that made the relevant state transitions.

As such, a number of important state transitions were not applied.

Additionally, missing from said upgrade handler were called to SetZoneAddressMapping() which allows for more efficient lookups of zone from ICAAddresses.

This PR updates the v1.4.5 upgrade handler that was not applied to be executed on v1.4.6, with the additional of calling SetZoneAddressMapping once for each of the ICA addresses controlled on each zone.

### DIFF
--- a/app/upgrades/types.go
+++ b/app/upgrades/types.go
@@ -23,6 +23,7 @@ const (
 	// mainnet upgrades
 	V010217UpgradeName = "v1.2.17"
 	V010405UpgradeName = "v1.4.5"
+	V010406UpgradeName = "v1.4.6"
 )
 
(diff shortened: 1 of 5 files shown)


faddat (approved):
Yeah unfortunately I see quite what happened here

faddat (approved):

PR #1069, opened 2024-01-21
Add an explicit build and test job for the icq relayer

faddat:
## 1. Summary

closes: #1067 

## 2.Type of change

Dependency update

## 3. Implementation details

* https://github.com/mvdan/github-actions-golang

## 4. How to test/use

This should be tested against traffic on mainnet to ensure that there are no problems.

## 5. Checklist

### DIFF
--- a/.github/workflows/icq-relayer-build-test.yml
+++ b/.github/workflows/icq-relayer-build-test.yml
@@ -1,6 +1,6 @@
 
 on: [push, pull_request]
-name: Test
+name: icq-relayer: build and test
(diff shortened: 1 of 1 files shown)


faddat:
@joe-bowman @tropicaldog I think I will likely just ninja-merge this, but if @tropicaldog wants to have a peek at it and a stab at review, that's cool. This is ci only.

faddat:
Def going to ninja merge this, so I can use it while working on #1067

PR #1086, opened 2024-01-24
Front end prod release

joe-bowman:
## 1. Summary
Fixes # (issue)

## 2.Type of change

## 5. Checklist


faddat:
checking through this now.

faddat (approved):
This is the squashed output of bunProd2 -> bunProdFinal (ran `find web-ui -not -path "web-ui/node_modules/*" -not -path "web-ui/.next/*" -type f -exec md5sum {} \; > out` against both branches, and the two out files are identical)

It in fact is exactly as described

PR #1142, opened 2024-02-13
auto add local denom for onboarded zone

### LINKED ISSUE #1121: Auto-add protocol data on onboarding zones
When a zone is registered, we should automatically add:
 - ConnectionProtocolData
 - LiquidAllowedDenomProtocolData (for the host zone; we cannot do other zones as we do not know the IBC denom for them ahead of time).

This means that tokens minted for a newly registered zone will be supported by claims from the first epoch after onboarding, without further governance.

joe-bowman:
## 1. Summary
Fixes #1121  - auto add local denom protocol data for newly registered zone.

## 2.Type of change

## 3. Implementation details

Protocol data must be added for all onboarded zones; however the protocol data for Connections and LocalDenom can, and should be automatically added on registration, as we have the details on hand to do so.

In the participation rewards AfterZoneCreated hook, we trigger the registration of LiquidAllowedDenomProtocolData for the Quicksilver chain, for the local qAsset denom.

This will allow claiming for local assets immediately from the first epoch after onboarding.

Note: the signature of the AfterZoneCreated() hook was changed to pass the `*Zone` instead of connectionId, ChainId and AccountPrefix to allow for greater functionality.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced consistency in zone creation handling across the system.
	- Improved data validation and storage mechanisms for zone creation events.
	- Removed unnecessary commented-out function for protocol data retrieval.
- **Tests**
	- Added a test function to verify zone creation behavior in the participation rewards keeper.

### DIFF
--- a/x/interchainstaking/keeper/proposal_handler.go
+++ b/x/interchainstaking/keeper/proposal_handler.go
@@ -101,7 +101,7 @@ func (k *Keeper) HandleRegisterZoneProposal(ctx sdk.Context, p *types.RegisterZo
 		return err
 	}
 
-	err = k.hooks.AfterZoneCreated(ctx, zone.ConnectionId, zone.ChainId, zone.AccountPrefix)
+	err = k.hooks.AfterZoneCreated(ctx, zone)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


faddat (approved):
elegance++:

* parameters reduced
* automation enhanced

PR #1172, opened 2024-02-21
fixes: #1169; add callback ID to query hash generation

### LINKED ISSUE #1169: Collision between two identical requests in the same block with different callbacks
## Summary of Bug

If we make two otherwise identical ICQ requests in the same block with a different callback, only the first will be respected. This is because we don't include the callback ID in the query hash. 

## Expected Behaviour

Two identical queries with different callback IDs ought to both be dispatched, as both callbacks need to be triggered

## Version

All

## Solution

Include callback ID in query key.

joe-bowman:
## 1. Summary
Fixes #1169 

This change adds the callbackID as part of the generated key used to identify hashes, This allows to otherwise identical requests to exist with different callbacks.

## 2.Type of change

## 3. Implementation details

Add an additional argument to GenerateQueryHash() allowing the inclusion of the callback ID in the key.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the query hash generation to include `callbackID`, improving query uniqueness and handling.
- **Refactor**
	- Updated query generation logic to incorporate `callbackID` in key generation for better query management.
- **Tests**
	- Modified tests to accommodate the inclusion of `callbackID` in the hash generation process and ensure robustness.

### DIFF
--- a/x/interchainquery/genesis_test.go
+++ b/x/interchainquery/genesis_test.go
@@ -70,7 +70,7 @@ func (s *InterChainQueryTestSuite) TestInitGenesis() {
 
 	interchainquery.InitGenesis(s.chainA.GetContext(), s.GetSimApp(s.chainA).InterchainQueryKeeper, types.GenesisState{Queries: []types.Query{*query}})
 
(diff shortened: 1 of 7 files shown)


faddat (approved):
if this PR: includes the callback ID in the query hash, 

then it looks great.

PR #1211, opened 2024-02-29
feat: add unbonding amount stats

### LINKED ISSUE #1123: Add unbonding info to zone stats
One the /zones endpoint, we generate statistics for each zone detailing deposits / delegated amount / supply / etc.

We should add unbonding statistics here, including:

`unbonding_amount`: Current number of tokens in the `unbonding` state.
`queued_amount`: Current number of tokens in the `queued` state.
`unbonding_count`: Number of withdrawal records in the `unbonding` state.
`queued_count`: Number of withdrawal records  in the `queued` state.
`unbond_record_count`: Number of active unbonding records.

Anything else that seems relevant.

tropicaldog:
## 1. Summary
Closes #1123

Todo: 

- [x] Add tests for new Keeper methods

## 2.Type of change

## 5. Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new fields in the `Statistics` message to provide detailed information on token amounts and withdrawal records in various states.
- **Enhancements**
	- Improved the calculation of unbonding amounts in the `GetRatio` function for more accurate data representation.
	- Added functionality to retrieve detailed information on unbonding and queued tokens, including counts.
- **Tests**
	- Updated and added test functions to ensure reliability of new unbonding and queuing functionalities.


joe-bowman:
Hey @tropicaldog - this is looking good. Just a couple of points:

1) conversion of function results to int32() and the functions already return int32s
2) the current design is very clean, but inefficient. Twice we have two methods iterating over the same set of keys in the KV store (twice for unbonding status, twice for queued status). Iterating over a store is expensive. I would suggest amending getUnbondingAmount() so that it counts _and_ sums, and where it is previously used, ignore the count; and change getQueuedAmount to do the same for the Queued records.
3) The unbonding record count is not zone specific. For each zone you will be outputting all unbonding records.

tropicaldog:
@joe-bowman what is the difference between WithdrawalRecord and UnbondingRecord? Is UnbondingRecord supposed to be a subset of WithdrawalRecord, with some unbond specific information?

joe-bowman:
Unbonding record represents an unbonding tx between the delegation ICA account and the validator on the host chain.

a withdrawal record represents an individual users’ withdrawal request.

depending on how withdrawals are allocated across validators, a withdrawal record may be associated with multiple unbonding records; but also an unbonding record may also cover multiple withdrawal records

joe-bowman (approved):
lgtm

faddat (approved):

PR #1223, opened 2024-03-05
fix(claimsmanager): check index to avoid runtime panic

### LINKED ISSUE #1217: x/claimsmanager/keeper: bytes.Index results should be checked before blindly assuming each key will always have 0x00,0x00 lest could cause a runtime panic from index out of bounds
## Summary of Bug
While auditing this module I noticed this code https://github.com/quicksilver-zone/quicksilver/blob/7966553ea6915777e87245a6c58ee467bc7ae979/x/claimsmanager/keeper/grpc_query.go#L73-L75 and https://github.com/quicksilver-zone/quicksilver/blob/7966553ea6915777e87245a6c58ee467bc7ae979/x/claimsmanager/keeper/grpc_query.go#L90-L92

Which assume that every single key will be properly formed and have the bytes in search. However, this could cause a runtime panic and take down the caller if at all we have any key that doesn't conform to that structure and could return a not found index. Sure +1 is always added to skip the separator but that's just gonna cause data corruption from the wrong result and say for example a key was of length 0, formed by one tampering with the database or innocently included due to some coding error somewhere, this will cause a runtime panic that could have been trivially avoided just by ensuring that bytes.Index returns a value >= 0

## Expected Behaviour
Firstly check that the index >= 0

/cc @joe-bowman @faddat 

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate contributors tagged/assigned

minhngoc274:
## 1. Summary
Fixes #1217 

## 2.Type of change

## 5. Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the robustness of index checking to prevent potential out-of-bounds errors.

### DIFF
--- a/x/claimsmanager/keeper/grpc_query.go
+++ b/x/claimsmanager/keeper/grpc_query.go
@@ -71,7 +71,7 @@ func (k Keeper) UserClaims(c context.Context, q *types.QueryClaimsRequest) (*typ
 		// check for the presence of the addr bytes in the key.
 		// first prefix byte is 0x00; so cater for that! Then + 1 to skip the separator.
 		idx := bytes.Index(key[1:], []byte{0x00}) + 1 + 1
-		if bytes.Equal(key[idx:idx+len(addrBytes)], addrBytes) {
+		if idx >= 0 && bytes.Equal(key[idx:idx+len(addrBytes)], addrBytes) {
 			out = append(out, claim)
 		}
 		return false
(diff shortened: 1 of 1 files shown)


faddat (approved):

joe-bowman (approved):
lgtm

odeke-em:
Am not sure though that this change is the right fix, I had prescribed that we check the value of bytes.Index not of bytes.Index + 1 + 1

x/claimsmanager/keeper/grpc_query.go
```diff
@@ -71,7 +71,7 @@ func (k Keeper) UserClaims(c context.Context, q *types.QueryClaimsRequest) (*typ
 		// check for the presence of the addr bytes in the key.
 		// first prefix byte is 0x00; so cater for that! Then + 1 to skip the separator.
 		idx := bytes.Index(key[1:], []byte{0x00}) + 1 + 1
```
odeke-em, line 73: We need to do the check here firstly before adding + 1 + 1
```go
idx := bytes.Index(key[1:], []byte{0x00})
if idx >= 0 {
     idx += 1 + 1 // Cater to skip the separator.
     if bytes.Equal(key[idx:idx+len(addrBytes)], addrBytes) {
         out = append(out, claim)
     }
}
```
```

PR #1234, opened 2024-03-05
fix(claimsmanager): properly check index to avoid invalid data in UserClaims,UserLastEpochClaims

### LINKED ISSUE #1217: x/claimsmanager/keeper: bytes.Index results should be checked before blindly assuming each key will always have 0x00,0x00 lest could cause a runtime panic from index out of bounds

odeke-em:
This change ensures that we correctly check for missing values in keeper.(UserClaims, UserLastEpochClaims) because otherwise due to the blind assumption that the key would always be well formed, +1 and +1+1 were being added respectively which would mean that the condition "idx >= 0" would ALWAYS pass and send over the wrong data if the key happened to contain the address bytes.

Fixes #1217
Supersedes PR #1223

## Summary by CodeRabbit

- **Refactor**
	- Improved the accuracy of claims data processing by adjusting key index calculations and byte comparisons.

### DIFF
--- a/x/claimsmanager/keeper/grpc_query.go
+++ b/x/claimsmanager/keeper/grpc_query.go
@@ -68,10 +68,17 @@ func (k Keeper) UserClaims(c context.Context, q *types.QueryClaimsRequest) (*typ
 	addrBytes := []byte(q.Address)
 
 	k.IterateAllClaims(ctx, func(_ int64, key []byte, claim types.Claim) (stop bool) {
+		// The assumption is that IterateAllClaims returns non-empty keys.
 		// check for the presence of the addr bytes in the key.
-		// first prefix byte is 0x00; so cater for that! Then + 1 to skip the separator.
-		idx := bytes.Index(key[1:], []byte{0x00}) + 1 + 1
-		if idx >= 0 && bytes.Equal(key[idx:idx+len(addrBytes)], addrBytes) {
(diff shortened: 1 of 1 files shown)


joe-bowman (approved):
Lgtm

joe-bowman (approved):

Issue #1200, opened 2024-02-27
Race condition with unbonding messages


joe-bowman:
We are able to unbond 7 times per unbonding period, hence the 3 day epoch (this works for most networks).

However, a race condition exists where the first unbonding to trigger for Epoch 7 is sent, before the final unbonding tx is completed for epoch, causing a failure.

We should observe the completion time of the unbond records for 7 epochs ago, and use this to trigger unbonding messages.

joe-bowman:
Closed by #1259

PR #1259, opened 2024-03-09
Fix low rr withdrawals

joe-bowman:
## 1. Summary
Makes good withdrawals that unbonded with below target RR, due to #1200.

- Looked back over the preceding 7 epochs, and calculated the correct RR `good_rr`.
- For all active unbondings, calculated `unredeemed_qatoms = burn_amount - (unbonding_amount / good_rr)` to determine how much of the original `burn_amount` was 'unredeemed` given that the unbonding amount represents a portion of the burn_amount at the `good_rr`
- mints the `unredeemed_rr` for each user, and sends to the unbonding escrow account.
- creates a withdrawal record for the `unredeemed` amount, in order to trigger the unbonding for the `unredeemed_amount` at the next epoch.

Note: it would have been cleaner to have removed the unbonding tokens from the existing records, but given we are already making changes to the requeued records in thus upgrade, it would make this overly complicated. Given the number of qAtoms is relatively small, re-minting the atoms is sufficient.

Note: also fix Makefile that got a newline added.

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected a formatting issue in the error message regarding the required Go version for building the application.
- **New Features**
	- Implemented a compensation mechanism for users affected by a redemption rate issue during specific epochs on cosmoshub-4. This feature automatically mints coins and creates new unbonding records for eligible users.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,8 +130,7 @@ BUILD_TARGETS := build install
 
 check_version:
 ifneq ($(GO_MINOR_VERSION),21)
-	@echo "ERROR: Go version 1.21 is required for building Quicksilver. Detected version: $(GO_MAJOR_VERSION).$(GO_MINOR_VERSION). There are 
-consensus breaking changes between binaries compiled with different Go versions."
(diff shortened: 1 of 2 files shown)


faddat (approved):
ah, that phrase, make whole

PR #1272, opened 2024-03-13
tolerate stale acks without panicking; fixes #1271

### LINKED ISSUE #1271: Stale ICA acknowledgements cause panic due to math.Int.Sub() resulting in negative balance
## Summary of Bug

When an issue has caused a packet acknowledgement to become stale, and its receipt causes a balance to be updated, we run the risk of a panic when the balance becomes negative (due to an ICQ request updating to the new value before the ack is received).

## Expected Behaviour

The transaction should handle the case where a negative balance is received, and complete gracefully without triggering any further state changes.

## Version

v1.5.0

## Steps to Reproduce

Attempt to deliver Juno channel-163 (icacontroller-juno-1.delegate) <--> channel-495 (icahost) sequence 226.

____

#### For Admin Use

- [ ] Not duplicate issue

joe-bowman:
## 1. Summary
Fixes #1271

## 2.Type of change

## 3. Implementation details

When we receive a stale ack that would make an account balance negative (this can happen with stale/delayed acks because the ICQ may update the account balance in the period between the ack being created and being delivered), we exit nil instead of panicking. We exit nil to ensure the packet is delivered, but we do not update the state futher.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of token redemption and delegation to prevent negative balances.
	- Added logging for better tracking of stale acknowledgments issues.

### DIFF
--- a/x/interchainstaking/keeper/ibc_packet_handlers.go
+++ b/x/interchainstaking/keeper/ibc_packet_handlers.go
@@ -938,7 +938,12 @@ func (k *Keeper) HandleRedeemTokens(ctx sdk.Context, msg sdk.Msg, amount sdk.Coi
 		}
 		k.Logger(ctx).Debug("outstanding delegations ack-received")
 		k.SetReceiptsCompleted(ctx, zone.ChainId, time.Unix(exclusionTimestampUnix, 0), ctx.BlockTime(), redeemMsg.Amount.Denom)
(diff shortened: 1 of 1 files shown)


minhngoc274 (commented):
LGTM

tropicaldog (approved):

PR #1319, opened 2024-03-22
Int64 to math

joe-bowman:
## 1. Summary
Fixes potential overflow when using 1e18 base tokens

Migrate a number of (u)int64 fields to sdkmath.Int

## 2.Type of change

## 3. Implementation details

Use sdkmath.Int in place of (u)int64 fields

## Summary by CodeRabbit

- **New Features**
	- Enhanced precision in financial calculations across the system by adopting a more robust mathematical framework.
- **Improvements**
	- Updated various data structures to support enhanced precision for amount fields.
	- Streamlined logic for financial operations, including distribution calculations and withdrawals.
- **Bug Fixes**
	- Fixed inconsistencies in financial data representation and calculations.
- **Refactor**
	- Simplified addition, subtraction, and comparison operations in financial calculations.
	- Enhanced validation logic for financial transactions to ensure accuracy and integrity.


tropicaldog (commented):
lgtm

x/interchainstaking/keeper/ibc_packet_handlers_test.go
```diff
@@ -350,37 +373,26 @@
...
-						BurnAmount: sdk.NewCoin("uqatom", sdk.NewInt(15000000)),
-						Txhash:     "d786f7d4c94247625c2882e921a790790eb77a00d0534d5c3154d0a9c5ab68f5",
-						Status:     types.WithdrawStatusQueued,
+						ChainId:      zone.ChainId,
+						Delegator:    addressutils.GenerateAccAddressForTest().String(),
+						Distribution: nil,
+						Recipient:    addressutils.GenerateAddressForTestWithPrefix(zone.AccountPrefix),
+						Amount:       nil,
+						BurnAmount:   sdk.NewCoin("uqatom", sdk.NewInt(15000000)),
+						Txhash:       "d786f7d4c94247625c2882e921a790790eb77a00d0534d5c3154d0a9c5ab68f5",
```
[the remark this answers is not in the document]
joe-bowman: This is a test transaction sha256 hash, not a key

x/interchainstaking/keeper/ibc_packet_handlers_test.go
```diff
@@ -217,39 +210,75 @@
...
```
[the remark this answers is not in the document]
joe-bowman: This is a test transaction sha256 hash, not a key

x/interchainstaking/keeper/ibc_packet_handlers.go
```diff
@@ -1443,9 +1443,15 @@ func (*Keeper) prepareRewardsDistributionMsgs(zone types.Zone, rewards sdkmath.I
 	}
 }
 
-func equalLsmCoin(valoper string, amount uint64, lsmAmount sdk.Coin) bool {
-	if strings.HasPrefix(lsmAmount.Denom, valoper) {
-		return lsmAmount.Amount.Equal(sdk.NewIntFromUint64(amount))
+func isNumericString(in string) bool {
+	_, err := strconv.Atoi(in)
+	return err == nil
+}
```
odeke-em, line 1449: In lieu of this change (using math.Int instead of uint64), let's add a comment here about why this code is okay: Given that if uqck is the smallest amount ever, maxInt64=((1<<63)-1)/1_000_000 = 9_223_372_036_854.775 (9.223 Trillion) but the highest supply of Quicksilver tokens in general supply is 400_000_000 (400 Million)

I highly recommend instead using strconv.ParseInt(in, 10, 64) and not strconv.Atoi because it fits within the value of bits so if on a 32-bit computer, it'll have a different range than on a 64-bit computer but strconv.ParseInt(..., 10, 64) will always check the range of 64-bits

```go
// It is okay to use strconv.ParseInt to test if a value is numeric because
// The total supply of QCK is 400_000_000 (400 million) QCK aka 400_000_000_000_000 UQCK
// and to parse numeric values, say in the smallest unit of uqck
// MaxInt64: (1<<63)-1 = 9_223_372_036_854_775_807 UQCK = 9_223_372_036_854.775 (9.223 Trillion) QCK
```

odeke-em: I've mailed out https://github.com/quicksilver-zone/quicksilver/pull/1322

joe-bowman: Thanks for catching this. The usage here is to validate the monotonically increasing value at the end of an lsm share. This is stored as a uint64 in the SDK logic (https://github.com/cosmos/cosmos-sdk/blob/0af2f4da004cbea6414a8bad56e8bdcd45badf1e/x/staking/keeper/tokenize_share_record.go#L14)

In reality it will never come close to 16 bits let alone 32, but happy to parseInt here.

PR #1322, opened 2024-03-22
x/interchainstaking/keeper: comment about isNumericString choices

odeke-em:
Adds a comment for future selves, about the choice of strconv.ParseInt inside isNumericString and showing that its limits won't be exceeded.

Updates PR #1319

## Summary by CodeRabbit

- **Refactor**
	- Improved numeric string validation in interchain staking logic to accommodate a wider range of token supplies.


joe-bowman (approved):
lgtm

PR #1339, opened 2024-03-26
fix(x/mint/simulation): use proper value of maxInt64

odeke-em:
This change uses the proper value of maxInt64, inferred from math.MaxInt64 and typed as an int64 then passed into math/rand.Rand.Int63n(maxInt64) instead of the non-standard prior fidgetting that set maxInt64 as an int(^uint(0)>>1) which sadly will overflow on 32-bit systems because the size of int on 32-bit systems is 32 bits, on 64-bit systems is 64 bits, and moreover the prior code used int64(Rand.Intn(maxInt64)) which is unnecessary.

## Summary by CodeRabbit

- **Refactor**
	- Improved the calculation of maximum integer values and randomized value generation in the minting process.

### DIFF
--- a/x/mint/simulation/genesis.go
+++ b/x/mint/simulation/genesis.go
@@ -5,6 +5,7 @@ package simulation
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"math/rand"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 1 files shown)


joe-bowman (approved):
lgtm - osmosis might appreciate the same change here: https://github.com/osmosis-labs/osmosis/blob/main/x/mint/simulation/genesis.go

faddat (approved):
definite improvement

odeke-em:
Done in https://github.com/osmosis-labs/osmosis/pull/7860, thanks for the reviews Jacob and Joe!

PR #1523, opened 2024-04-30
fix(icq-relayer): make "keys list" print out keys in a deterministic order

### LINKED ISSUE #1522: icq-relayer: keys list command prints out chain associated keys in a non-deterministic order
## Summary of Bug
The code below https://github.com/quicksilver-zone/quicksilver/blob/7f095c3614cf580d2bf6f7bff3ac7b34bff7644a/icq-relayer/cmd/keys.go#L197-L200 when ran prints out keys in a non-deterministic order each time as it iterates over a map.

This can be annoying for integrity checks that might want to checksum and detect for added keys without having to do special filtration and sorting each time for which the remedy currently is
```shell
$ icq-relayer keys list ibc-0 | sort | uniq
```

## Expected Behaviour
A deterministic order of keys

## Version
7f095c3614cf580d2bf6f7bff3ac7b34bff7644a

I have a PR coming up!

#### For Admin Use

- [x] Not duplicate issue

odeke-em:
This ensures that new entries or removals can easily be tailed or detected through simple checksumming.

Fixes #1522


faddat (approved):
I could see this being useful for integrations one day.

PR #1525, opened 2024-04-30
fix(icq-relayer/config): return immediately on non-os.IsNotExist errors

### LINKED ISSUE #1524: icq-relayer: initConfig invokes os.Stat and presumes that the only error returned is NotExists and brute-force invokes config.CreateConfig
## Summary of Bug
Noticed while auditing that there is this code https://github.com/quicksilver-zone/quicksilver/blob/7f095c3614cf580d2bf6f7bff3ac7b34bff7644a/icq-relayer/cmd/config.go#L25-L31
which on any error just falls back to config.CreateConfig which brute-force tries to create the configuration.

However, there can be simply a permission error or a device/drive linking error (which can occur on Cloud computing environments say with a persistent disk not being mounted correctly or in containers)

## Expected Behaviour
We should have a correct switch of errors whereby all other errors except os.IsNotExist are returned immediately.

## Version
7f095c3614cf580d2bf6f7bff3ac7b34bff7644a

/cc @faddat @joe-bowman 
____

#### For Admin Use

- [x] Not duplicate issue

odeke-em:
This change avoids the need to try to brute-force creating an entirely fresh configuration, in the case where an error that isn't a path non-existence error occurs, for example in the case of a permission error on the parent directory or non-matching user/group permissions. Now we check against the returned error after os.Stat and accordingly return immediately or continue to create the configuration.

Fixes #1524

### DIFF
--- a/icq-relayer/cmd/config.go
+++ b/icq-relayer/cmd/config.go
@@ -24,11 +24,15 @@ func initConfig(cmd *cobra.Command) error {
 	cfgPath := path.Join(home, "config.yaml")
 	_, err = os.Stat(cfgPath)
 	if err != nil {
-		err = config.CreateConfig(home)
-		if err != nil {
+		if !os.IsNotExist(err) { // Return immediately
+			return err
+		}
+
+		if err := config.CreateConfig(home); err != nil {
 			return err
 		}
 	}
+
 	viper.SetConfigFile(cfgPath)
 	err = viper.ReadInConfig()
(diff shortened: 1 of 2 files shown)


tropicaldog (dismissed):
LGTM. Thanks @odeke-em for the PR!

tropicaldog (approved):

joe-bowman (approved):
lgtm

PR #1533, opened 2024-05-01
Fix deprecated rule in .golangci.yml

pysel:
## 1. Summary
Fixes #None

## 2.Type of change

## 5. Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,8 +2,6 @@ run:
   tests: true
   #   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 10m
-  skip-dirs:
-    - third-party-chains
 
 linters:
   disable-all: true
@@ -43,6 +41,8 @@ issues:
 
   max-issues-per-linter: 10000
   max-same-issues: 10000
+  exclude-dirs:
+    - third-party-chains
 
 linters-settings:
   gci:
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -2,8 +2,6 @@ run:
   tests: true
   #   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 10m
-  skip-dirs:
-    - third-party-chains
```
pysel, line 6:
```
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`. 
```

pysel:
```
ruslanakhtariev@169-231-72-96 quicksilver % golangci-lint version                         
golangci-lint has version 1.57.2 built with go1.22.1 from 77a8601a on 2024-03-28T19:01:11Z
```

faddat (approved):

faddat (changes requested):
Hey, it seems that this has strangely angered the winter. Could you check out what the issue is?

joe-bowman (approved):
lgtm

joe-bowman:
I believe the issue here is that .github/workflows/golangci.yaml is using 1.56 not 1.57 and the setting while valid for your local instance, is not valid for the CI. Updating the CI version to 1.57.2 should do the trick, thank you!

PR #1576, opened 2024-05-07
Add workflow that checks stale PRs 

pysel:
## 1. Summary
Fixes # (issue)

## 2.Type of change

Workflow that marks PRs as stale if they received no activity for 14 days, closes if no further action is taken after 6 days.

Can be disabled on specific PRs if label `keepalive` is attached


faddat:
I'm not sure that we'd actually want this here, some PR's go for longer than that. 

@joe-bowman -- what do you think?

joe-bowman:
I'm not fundamentally against it - no activity for 20 days is pretty long tbh. They can be reopened, or tagged if they need to be inactive for a period.

pysel:
> I'm not sure that we'd actually want this here, some PR's go for longer than that.
> 
> @joe-bowman -- what do you think?

if a PR should be open for a long time without any activity, a `keepalive` label can be attached to it in order to keep it open indefinitely

cc: @faddat @joe-bowman

.github/workflows/stalebot.yaml
```diff
@@ -0,0 +1,27 @@
...
+          repo-token: ${{ secrets.GH_PAT }}
+          stale-pr-message: >
+            This pull request has been automatically marked as stale because it
+            has not had any recent activity. It will be closed if no further
+            activity occurs. Thank you!
+          days-before-stale: -1
+          days-before-close: -1
+          days-before-pr-stale: 14
+          days-before-pr-close: 6
+          exempt-pr-labels: "security, proposal, blocked, keepalive"
```
pysel, line 27:
```suggestion
          exempt-pr-labels: "keepalive"
```

joe-bowman (approved):
Lfg

PR #1587, opened 2024-05-07
remove relative reference to quicksilver in icq-relayer

joe-bowman:
remove relative reference to quicksilver in icq-relayer, as this regularly breaks dependency upgrades


faddat:
Oh, I like this one!

faddat (approved):

PR #1592, opened 2024-05-07
add support for price calc from CL pools, multiple pools, plus multi-hop

### LINKED ISSUE #1357: Refactor relative token price logic
## Summary

We currently determine the token price of each baseDenom in the participation rewards module on the fly, because it is only used for determining relative tvls for the purpose of participation rewards allocation. However, we now also require this logic for dynamically determining the dust threshold (based on dollar amounts), so we should extract the current logic, add osmo->usdc pricing logic, and store the usd value of each token for use by other modules.

joe-bowman:
## 1. Summary
Partially fixes #1357

- Adds support to use CL pools for price determination
- Add support for averaging across multiple pools
- Add multi-hop support where base-denom <-> token pair doesn't exist

1) Create a slice of prices for each denom0 <-> denom1 pair
2) use Depth First Search to determine a price for base_denom <-> denom pair, using intermediate if there is no direct paring


faddat:
....this is epic!

Going to give it a deep read through, but also want to note what I think is so epic about it -- I'm not aware of any other code that does this on a chain <-> chain basis.

faddat (approved):

PR #896, opened 2023-12-12
perf: hoist regexp into globals

joe-bowman:
## 1. Summary
Fixes #804 

Move regexp compiles to global vars for perf reasons.

## 2.Type of change

- [x] Performance fix (non-breaking change, no functional change)

Issue #804, opened 2023-11-27
all, x/interchainstaking/types: hoist regexp made from constants into globals to massively improve performance in IntentsFromString


odeke-em:
## Summary of Bug
If we read this code inside https://github.com/quicksilver-zone/quicksilver/blob/aa4f358697bf2ce4f5c0b40f793232b83f81d70f/x/interchainstaking/types/msgs.go#L92C6-L92-L101 we can see that to match validity of each intent we have to construct 2 regular expressions each time.
Let's note that those regexps are constructed from a static string and we are wasting CPU cycles and RAM whenever we run through that code

## Exhibit
If we use this benchmark code
```go
package types_test

import (
	"testing"

	sdk "github.com/cosmos/cosmos-sdk/types"
	"github.com/google/go-cmp/cmp"

	"github.com/quicksilver-zone/quicksilver/x/interchainstaking/types"
)

var benchSink any = nil

func BenchmarkIntentsFromString(b *testing.B) {
	str := "0.3cosmosvaloper1sjllsnramtg3ewxqwwrwjxfgc4n4ef9u2lcnj0,0.3cosmosvaloper156gqf9837u7d4c4678yt3rl4ls9c5vuursrrzf,0.4cosmosvaloper1a3yjj7d3qnx4spgvjcwjq9cw9snrrrhu5h6jll"

	want := []*types.ValidatorIntent{
		{
			ValoperAddress: "cosmosvaloper1sjllsnramtg3ewxqwwrwjxfgc4n4ef9u2lcnj0",
			Weight:         sdk.MustNewDecFromStr("0.3"),
		},
		{
			ValoperAddress: "cosmosvaloper156gqf9837u7d4c4678yt3rl4ls9c5vuursrrzf",
			Weight:         sdk.MustNewDecFromStr("0.3"),
		},
		{
			ValoperAddress: "cosmosvaloper1a3yjj7d3qnx4spgvjcwjq9cw9snrrrhu5h6jll",
			Weight:         sdk.MustNewDecFromStr("0.4"),
		},
	}

	b.ReportAllocs()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		got, err := types.IntentsFromString(str)
		if err != nil {
			b.Fatal(err)
		}
		if diff := cmp.Diff(got, want); diff != "" {
			b.Fatalf("Mismatch: got - want +\n%s", diff)
		}
		benchSink = got
	}

	if benchSink == nil {
[...]

PR #1637, opened 2024-05-29
test(x/interchainstaking/types): add BenchamrkIntentsFromString + minor perf

odeke-em:
Following up with PR #896 from performance advisory #804, this change brings in the benchmark and also adds a minor performance improvement to further hoist into a global the cosmos-sdk.Dec value of 1, and also pre-allocate using a size hint.

Updates #804

### DIFF
--- a/x/interchainstaking/types/bench_test.go
+++ b/x/interchainstaking/types/bench_test.go
@@ -1,0 +1,117 @@
+package types_test
+
+import (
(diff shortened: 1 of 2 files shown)


joe-bowman:
Can you lint this @odeke-em so we can merge it please?

joe-bowman:

PR #1680, opened 2024-06-29
test: fix v1.6.1 upgrade test

tropicaldog:
## 1. Summary
Fix the upgrade test failed to run due to wrong chain-id.

## 2.Type of change

## 5. Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured upgrades are correctly applied only on the production chain by setting the chain ID during the upgrade process.

### DIFF
--- a/app/upgrades_test.go
+++ b/app/upgrades_test.go
@@ -896,9 +896,9 @@ func (s *AppTestSuite) TestV010505UpgradeHandler() {
 func (s *AppTestSuite) TestV010601UpgradeHandler() {
 	s.InitV160TestZones()
 	app := s.GetQuicksilverApp(s.chainA)
-
 	ctx := s.chainA.GetContext()
-
+	// upgrade only run on production chain
+	ctx = ctx.WithChainID(upgrades.ProductionChainID)
 	handler := upgrades.V010601UpgradeHandler(app.mm,
 		app.configurator, &app.AppKeepers)
 


app/upgrades_test.go
```diff
@@ -896,9 +896,9 @@ func (s *AppTestSuite) TestV010505UpgradeHandler() {
 func (s *AppTestSuite) TestV010601UpgradeHandler() {
 	s.InitV160TestZones()
 	app := s.GetQuicksilverApp(s.chainA)
-
 	ctx := s.chainA.GetContext()
-
+	// upgrade only run on production chain
+	ctx = ctx.WithChainID(upgrades.ProductionChainID)
```
joe-bowman, line 901: A better fix would be to make the IsMainnet() check in the upgrade to a IsMainnet() || IsTest()

tropicaldog: @joe-bowman I guess adjusting the test for local, testnet or mainnet a better pattern? In my opinion upgrade handler should not take the test into account.

TropicalDog17: @joe-bowman but the way we setup the chain, the chainID would not be testnet1, thus isTest() will return false

odeke-em:
Kind ping @joe-bowman @tropicaldog as this PR has stalled other PRs and CI being green :-)

app/upgrades_test.go
tropicaldog: Hey @joe-bowman IsTest() won't work since the chain id here is not `testnet2` but some chain-id provided by the ibc test suite. I think my solution/workaround is good for now.

joe-bowman:
This is resolved in #1696 - the fix here doesn't actually test the upgrade handler.

PR #1712, opened 2024-08-15
feat: add force ica send governance message type

### LINKED ISSUE #1706: Add Governance message type to force transfer from an ICA account
## Summary

Due to a closed ICA channel, it can happen that a user makes a deposit, ICQ finds and recognises their deposit and correct allocates the qAsset pertaining to the deposit, but the subsequent ICA MsgSend from the DepositICA to the DelegateICA fails. There is no retry mechanism here, as we rely on the deposit account having a non-zero balance to trigger the deposit flow, so we must tie the deposit/receipt/mint flow with the send from DepositICA to DelegateICA. 

This appears to have happened twice on ssc-1. Two deposits we made, recognised and the qSaga minted, but the deposit was never moved into the delegate account and delegated.

A governance message type should exist to forceTransfer from the depositAccount to another account (I can't see why this should be anything other than the delegate account, but this seems to restrictive - I feel we can safely move assets between any of the ICA accounts safely).

**MsgICASend**

**srcAccount** (string): port name of the source account e.g.: icacontroller-ssc-1.deposit
**dstAccount** (string): port name of the destination account e.g: icacontroller-scc-1.delegate
**amount** (sdk.Coins): amount to transfer
**fromAddress**: MUST be the governance account

joe-bowman:
## 1. Summary
Fixes #1706 

## 2.Type of change

## 3. Implementation details

Add a new governance type that permits funds to be MsgSend from an ICA account to another account, subject to the following constraints.
- Source account must be a valid ICA. 
- Destination account must be a valid account on the same network.
- From must be the governance authority. 

TODO:
- Tests


joe-bowman:

joe-bowman:

joe-bowman:

joe-bowman:

PR #1737, opened 2024-11-07
Release/v1.7.0-rc1

joe-bowman:
## 1. Summary
Fixes a number of small bugs:
 - ensure we register inclusion proof implementations
 - guard against panic in claimsmanager hook
 - handle local zone containing multiple hyphens
 - fix bug with share proof reconstruction if non-zero starting index
 - handle compatibility mode config in icq-relayer
 - properly handle tendermint tx proofs in icq-relayer (introduced in v1.1.0)


ajansari95 (approved):
lgtm

PR #1788, opened 2025-01-10
Add claimsmanager udpate spec.

arhamchordia:


x/claimsmanager/spec/README.md
```diff
@@ -129,11 +129,107 @@ type Proof struct {
...
+message MsgSubmitClaim {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+  string module_address = 1 [ json_name = "module_address", (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  string event_name = 2 [ json_name = "zone" ];
+  // todo add more fields
+}
+```
+
+### MsgDeleteClaimableEvent
```
joe-bowman, line 204: Do we require this?

I think just having a method on the Keeper to do this is sufficient. It should be called by whichever module created the event in the first place. Should not require a tx to remove anything.

arhamchordia: Makes sense. I will remove it then. 

My idea about the design is to be controlled by contracts as well.

joe-bowman: That is a fair point. I’m not terribly familiar with how bindings between cw and go logic work.

arhamchordia: Contracts need proto messages that they can call. 
While writing rust contracts we need a QuickSilver proto lib that gives rust code for calling those endpoints.

proto/quicksilver/claimsmanager/v1/messages.proto
```diff
@@ -1,7 +1,59 @@
...
+  string zone = 2 [json_name = "zone"];
+  string src_zone = 3 [json_name = "src_zone"];
+  quicksilver.claimsmanager.v1.ClaimType claim_type = 4 [json_name = "claim_type"];
+  repeated quicksilver.claimsmanager.v1.Proof proofs = 5 [json_name = "proofs"];
+}
+
+// MsgSubmitClaimResponse defines the MsgSubmitClaim response type.
+message MsgSubmitClaimResponse {}
+
+message MsgSubmitClaimableEventClaim {
```
arhamchordia, line 46: This is not needed. Will get rid of it later.

joe-bowman:

joe-bowman:

PR #1866, opened 2025-04-23
V1.8.0 release

joe-bowman:
## 1. Summary
Adds upgrade handler for v1.8.0 that removes two stuck withdrawal records, that were both paid out on juno before the channel closed, so no ack was received.

It also removes juno unbonding_records for epoch 277 that were also unacked but recreated in e278, and are now redundant.

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved issues with extra or duplicated unbonding and withdrawal records for the "juno-1" chain during the upgrade process.
- **Tests**
	- Added tests to verify correct cleanup of unbonding and withdrawal records during the upgrade.
- **Chores**
	- Updated and added several dependencies to improve reliability and maintain compatibility.


ajansari95 (dismissed):
LGTM

PR #2049, opened 2026-04-15
fix(upgrade): v1.10.2 mint-to-cover sunset (patch for halted chain)

qbyt2:
The original v1.10.2 sunset of stargaze-1 and omniflixhub-1 panicked at upgrade height 17,775,700: the escrow balance was ~1,387 qSTARS short of the outstanding qSTARS refund obligation, causing SendCoinsFromModuleToAccount to fail partway through the stargaze refund loop. Since upgrades execute under the ABCI cache multi-store, the panic aborted the upgrade and halted the chain.

This patch replaces the v1.10.2 handler with a mint-to-cover strategy: before refunding, it compares escrow balance against total obligation; if short, it mints the exact shortfall into the interchainstaking module account and transfers it to the escrow module before refunding. The minted qAssets are economically backed by the base-denom balances held across the zone's ICA accounts on the host chain, which will be reclaimed via a post-sunset governance action.

The handler also carries a defensive guard that skips WDRs with invalid or zero BurnAmount (legacy pre-v1.10.1 records), so sdk.NewCoins cannot panic mid-loop and the mint branch cannot over-mint against obligations we can't honor.

Upgrade handler is registered under the same "v1.10.2" name so the stuck on-chain plan at height 17,775,700 resumes automatically once validators run the patched binary — no --unsafe-skip-upgrades flag or new gov prop required.

Tests cover: shortfall (mint path), exact match (no mint), surplus (no mint), no records, zone not found, and malformed-WDR skipping.

## 1. Summary
Fixes # (issue)

## 2.Type of change

## 5. Checklist

### DIFF
--- a/app/upgrades/upgrades.go
+++ b/app/upgrades/upgrades.go
@@ -31,7 +31,7 @@ func Upgrades() []Upgrade {
 		// v1.10.1 - HandleFailedUndelegate BurnAmount zero guard hotfix
 		{UpgradeName: V0101001UpgradeName, CreateUpgradeHandler: NoOpHandler},
 
-		// v1.10.2 - Sunset stargaze & omniflix + HandleFailedUndelegate fix
+		// v1.10.2 - Zone sunset (stargaze-1, omniflixhub-1) with mint-to-cover refund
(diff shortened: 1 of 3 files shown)


joe-bowman (approved):
lgtm

